PR #12, opened 2021-09-03
Add set for Laravel 8.x

zingimmick:
- [x] Add argument
  - [x] The `castUsing` method of the `Illuminate\Contracts\Database\Eloquent\Castable` interface has been updated to accept an array of arguments. https://github.com/laravel/framework/commit/4d228d6e9dbcbd4d97c45665980d8b8c685b27e6
- [x] Add default value
  - [x] The `listen` method of the `Illuminate\Contracts\Events\Dispatcher` contract has been updated to make the `$listener` property optional. https://github.com/laravel/framework/commit/46084d946cdcd1ae1f32fc87a4f1cc9e3a5bccf6
- [x] Rename property
  - [x] The previously deprecated `$app` property of the `Illuminate\Support\Manager` class has been removed. https://github.com/laravel/framework/pull/32092
  - [x] For consistency with other features of Laravel, the `retryAfter` method and `retryAfter` property of queued jobs, mailers, notifications, and listeners have been renamed to `backoff`. https://github.com/laravel/framework/commit/4656c2cf012ac62739ab5ea2bce006e1e9fe8f09
  - [x] The `timeoutAt` property of queued jobs, notifications, and listeners has been renamed to `retryUntil`. https://github.com/laravel/framework/commit/12c35e57c0a6da96f36ad77f88f083e96f927205
- [x] Rename method
  - [x] The previously deprecated `sendNow` method has been removed. Instead, please use the `send` method. https://github.com/laravel/framework/pull/32092
  - [x] For consistency with other features of Laravel, the `retryAfter` method and `retryAfter` property of queued jobs, mailers, notifications, and listeners have been renamed to `backoff`. https://github.com/laravel/framework/commit/4656c2cf012ac62739ab5ea2bce006e1e9fe8f09
  - [x] The `timeoutAt` property of queued jobs, notifications, and listeners has been renamed to `retryUntil`. https://github.com/laravel/framework/commit/12c35e57c0a6da96f36ad77f88f083e96f927205
  - [x] The `decodeResponseJson` method that belongs to the `Illuminate\Testing\TestResponse` class no longer accepts any arguments. https://github.com/laravel/framework/commit/f9374fa5fb0450721fb2f90e96adef9d409b112c
  - [x] The `assertExactJson` method now requires numeric keys of compared arrays to match and be in the same order. https://github.com/laravel/framework/commit/fd662d4699776a94e7ead2a42e82c340363fc5a6
- [x] If your `App\Providers\EventServiceProvider` class contains a register function, you should ensure that you call `parent::register` at the beginning of this method. https://github.com/laravel/framework/commit/f1289515b27e93248c09f04e3011bb7ce21b2737
- [x] For consistency with other dispatching methods, the `allOnQueue()` and `allOnConnection()` methods used with job chaining have been removed. You may use the `onQueue()` and `onConnection()` methods instead. https://github.com/laravel/framework/commit/de662daf75207a8dd69565ed3630def74bc538d3

### DIFF
--- a/src/Rector/ClassMethod/AddArgumentDefaultValueRector.php
+++ b/src/Rector/ClassMethod/AddArgumentDefaultValueRector.php
@@ -1,0 +1,117 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Laravel\Rector\ClassMethod;
+
+use PhpParser\BuilderHelpers;
+use PhpParser\Node;
+use PhpParser\Node\Param;
+use PhpParser\Node\Stmt\ClassMethod;
+use Rector\Core\Contract\Rector\ConfigurableRectorInterface;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Laravel\ValueObject\AddArgumentDefaultValue;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\ConfiguredCodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+use Webmozart\Assert\Assert;
+
+/**
+ * @see \Rector\Laravel\Tests\Rector\ClassMethod\AddArgumentDefaultValueRector\AddArgumentDefaultValueRectorTest
+ */
+final class AddArgumentDefaultValueRector extends AbstractRector implements ConfigurableRectorInterface
+{
+    /**
+     * @var string
+     */
+    public const ADDED_ARGUMENTS = 'added_arguments';
+
+    /**
+     * @var AddArgumentDefaultValue[]
+     */
+    private array $addedArguments = [];
+
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition(
+            'Adds default value for arguments in defined methods.',
+            [
+                new ConfiguredCodeSample(
+                    <<<'CODE_SAMPLE'
+class SomeClass
+{
+    public function someMethod($value)
+    {
+    }
+}
+CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
+class SomeClass
+{
+    public function someMethod($value = false)
+    {
+    }
+}
+CODE_SAMPLE
+                    ,
+                    [
+                        self::ADDED_ARGUMENTS => [
+                            new AddArgumentDefaultValue('SomeClass', 'someMethod', 0, false),
+                        ],
+                    ]
+                ),
+            ]
+        );
+    }
+
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [ClassMethod::class];
+    }
+
+    /**
+     * @param ClassMethod $node
+     */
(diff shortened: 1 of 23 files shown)


TomasVotruba:
Thank you :+1:

src/Rector/ClassMethod/AddArgumentDefaultValueRector.php
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * @param array<string, AddArgumentDefaultValue[]> $configuration
+     */
+    public function configure(array $configuration): void
+    {
+        $addedArguments = $configuration[self::ADDED_ARGUMENTS] ?? [];
+        Assert::allIsInstanceOf($addedArguments, AddArgumentDefaultValue::class);
+        $this->addedArguments = $addedArguments;
+    }
+}
```
TomasVotruba, line 117: Just noticted this... maybe you can use: https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#argumentadderrector

zingimmick: I tried `ArgumentAdderRector` before, but it only works when there is no argument in the position. The problem here is that there is already an argument without default value in this position, and it needs to be optional. 

Here is the configuration that didn't work before:
```php
$services->set(ArgumentAdderRector::class)
    ->call('configure', [[
        ArgumentAdderRector::ADDED_ARGUMENTS => ValueObjectInliner::inline([
            new ArgumentAdder('Illuminate\Contracts\Events\Dispatcher', 'listen', 1, 'listener', null),
        ]),
    ]]);
```

The option `added_arguments` here seems to be ambiguous.
https://github.com/rectorphp/rector-laravel/blob/b58ac1f66ac9723b5a2eb4b5d7a218d0404d1b47/src/Rector/ClassMethod/AddArgumentDefaultValueRector.php#L23-L26

TomasVotruba: > I tried ArgumentAdderRector before, but it only works when there is no argument in the position. 

I see. This feature should be supported with that rule too. E.g. with `*ANY*` constant in this rule:
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#replaceargumentdefaultvaluerector

The split packages like rector-laravel should only contain Laravel-specific rules. This can be re-used in generic way, so it should located in rector-src. Could you move it there?

zingimmick: > I see. This feature should be supported with that rule too.

Created a draft PR in case I misunderstood what you mean rectorphp/rector-src#837.

> E.g. with `*ANY*` constant in this rule:
> https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#replaceargumentdefaultvaluerector

I'm not sure about the scope of this constant. `ReplaceArgumentDefaultValueRector` only handles `MethodCall` and `StaticCall` now, replace any value before with value after?

PR #13, opened 2021-09-07
Add set for Laravel legacy factories to classes

zingimmick:
### Definition
Before: https://laravel.com/docs/7.x/database-testing#writing-factories
After: https://laravel.com/docs/8.x/database-testing#defining-model-factories
### Usage
Before: https://laravel.com/docs/7.x/database-testing#using-factories
After: https://laravel.com/docs/8.x/database-testing#creating-models-using-factories

### Tasks
- [x] Convert factory definition to class.
- [x] Use the static factory method instead of global factory function.
- [x] Applying states by specify the name of each state to simply call the state methods

### Missing jobs:
- Remove `classmap` block from the `autoload` section and add the new namespaced class directory mappings in the composer.json file.
- Add the `Database\Factories` namespace to those classes.

I think the `NormalizeNamespaceByPSR4ComposerAutoloadRector` can handle these missing jobs well, so these jobs have not been processed for the time being.


src/Rector/Namespace_/FactoryDefinitionRector.php
```diff
@@ -0,0 +1,228 @@
...
+            $this->processFactoryConfiguration($factories[$name->toString()], $stmt->expr);
+
+            unset($node->stmts[$key]);
+        }
+
+        foreach ($factories as $factory) {
+            $node->stmts[] = $factory;
+        }
+
+        return $node;
```
samsonasik, line 121: should return null when `$factories` is empty: 

```suggestion
        if ($factories === []) {
            return null;
        }

        $node->stmts = array_merge($node->stmts, $factories);
        return $node;
```

src/Rector/Namespace_/FactoryDefinitionRector.php
```diff
@@ -0,0 +1,228 @@
...
+        $this->appendAfterCalling($class, $methodCall, $name);
+    }
+
+    private function addDefinition(Class_ $class, MethodCall $methodCall): void
+    {
+        if (count($methodCall->args) !== 2) {
+            return;
+        }
+
+        $callback = $methodCall->args[1]->value;
```
samsonasik, line 192: need to have `isset($methodCall->args[1])` check to skip when index removed by other rules.

src/Rector/Namespace_/FactoryDefinitionRector.php
```diff
@@ -0,0 +1,228 @@
...
+        $class->stmts[] = $this->modelFactoryNodeFactory->createStateMethod($methodCall);
+    }
+
+    private function appendAfterCalling(Class_ $class, MethodCall $methodCall, string $name): void
+    {
+        if (count($methodCall->args) !== 2) {
+            return;
+        }
+
+        $closure = $methodCall->args[1]->value;
```
samsonasik, line 215: need to have `isset($methodCall->args[1])` check to skip when index removed by other rules.

src/Rector/FuncCall/FactoryFuncCallToStaticCallRector.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        if (! $this->isName($node->name, self::FACTORY)) {
+            return null;
+        }
+
+        if ($node->args === []) {
+            return null;
+        }
+
+        $firstArgValue = $node->args[0]->value;
```
samsonasik, line 65: need to have `isset($node->args[0])` check to avoid index removed by other rules.

src/Rector/FuncCall/FactoryFuncCallToStaticCallRector.php
```diff
@@ -0,0 +1,80 @@
...
+
+        $model = $firstArgValue->class;
+
+        // create model
+        if (count($node->args) === 1) {
+            return new StaticCall($model, self::FACTORY);
+        }
+
+        // create models of a given type
+        return new StaticCall($model, self::FACTORY, [$node->args[1]]);
```
samsonasik, line 78: need to have `isset($node->args[1])` check to avoid index removed by other rules.

zingimmick: All isset check added.

src/Rector/MethodCall/FactoryApplyingStatesRector.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * @param Arg[] $args
+     * @return mixed[]
+     */
+    private function getStatesFromArgs(array $args): array
+    {
+        if (count($args) === 1 && isset($args[0])) {
+            return (array) $this->valueResolver->getValue($args[0]->value);
```
samsonasik, line 79: Latest nikic/php-parser can have Arg in both types: Arg, and VariadicPlacehoder that doesn't has value property, see

https://github.com/nikic/PHP-Parser/blob/50953a2691a922aa1769461637869a0a2faa3f53/lib/PhpParser/Node/Expr/FuncCall.php#L12

please add check arg instanceof Arg check for it

zingimmick: Ref: https://github.com/rectorphp/rector-src/pull/904, I'll fix it after this PR merge.

TomasVotruba:
Let me know when this is ready to merge :+1:

src/Rector/MethodCall/FactoryApplyingStatesRector.php
zingimmick: > Latest nikic/php-parser can have Arg in both types: Arg, and VariadicPlacehoder that doesn't has value property, see
> 
> https://github.com/nikic/PHP-Parser/blob/50953a2691a922aa1769461637869a0a2faa3f53/lib/PhpParser/Node/Expr/FuncCall.php#L12
> 
> please add check arg instanceof Arg check for it

@samsonasik All checks have passed again on nikic/php-parser 4.13.0.

samsonasik (approved):

zingimmick:
> Let me know when this is ready to merge 👍

@TomasVotruba If no other changes are required, I think it can be merged now.

TomasVotruba:
Merged 👍

zingimmick:
> Merged 👍

Thank you 👍

PR #15, opened 2021-09-09
Add RouteActionCallableRector

zingimmick:
Use PHP callable syntax instead of string syntax for controller route declarations, which provides better support for jumping to the controller class in many IDEs.

```diff
use Illuminate\Support\Facades\Route;

-Route::get('/users', 'App\Http\Controllers\UserController@index');
+Route::get('/users', [\App\Http\Controllers\UserController::class, 'index']);
```

### DIFF
--- a/src/Rector/StaticCall/RouteActionCallableRector.php
+++ b/src/Rector/StaticCall/RouteActionCallableRector.php
@@ -103,7 +103,7 @@ CODE_SAMPLE
             return null;
         }
 
-        [$controller,$method] = $segments;
+        [$controller, $method] = $segments;
         $namespace = $this->getNamespace($this->file->getSmartFileInfo());
         if (! str_starts_with($controller, '\\')) {
             $controller = $namespace . '\\' . $controller;
--- a/src/Rector/StaticCall/RouteActionCallableRector.php
+++ b/src/Rector/StaticCall/RouteActionCallableRector.php
@@ -10,10 +10,10 @@ use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Identifier;
 use PHPStan\Reflection\Php\PhpMethodReflection;
-use PHPStan\Type\ObjectType;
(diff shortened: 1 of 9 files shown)


zingimmick:
All checks have passed 🎉 @TomasVotruba it is ready for review.

src/Rector/StaticCall/RouteActionCallableRector.php
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        $arg = $node->args[$position];
+
+        $action = $this->valueResolver->getValue($arg->value);
+        if (! $this->isActionString($action)) {
+            return null;
+        }
+
+        [$controller, $method] = explode('@', $action);
```
samsonasik, line 101: while technically never met, there should be a check for multiple `@`, eg: `Foo@bar@baz`, .so only applicable when string only have 1 `@` in the middle.

zingimmick: Thanks for reminding, added a test fixture for that.

src/Rector/StaticCall/RouteActionCallableRector.php
```diff
@@ -0,0 +1,184 @@
...
+
+        if ($node instanceof MethodCall && $this->nodeTypeResolver->isObjectTypes(
+            $node->var,
+            [new ObjectType('Illuminate\Routing\Router'), new ObjectType('Illuminate\Routing\RouteRegistrar')]
+        )) {
+            return true;
+        }
+
+        return $node instanceof StaticCall && $this->isName($node->class, 'Illuminate\Support\Facades\Route');
+    }
```
TomasVotruba, line 152: This looks like complex re-usable logic. Could you extact is to own service?

zingimmick: Extract it to `RouterRegisterNodeAnalyzer`.

samsonasik (approved):

TomasVotruba:
Thank you :+1:

PR #16, opened 2021-09-26
Add UnifyModelDatesWithCastsRector

zingimmick:
Ref: https://github.com/laravel/framework/pull/32856

Unifying `$dates` with `$casts` will remove ambiguity and give developers a single place for casting their _date_ attributes as well as other attributes. 

**Before**

```php
class Person extends Model
{
    protected $casts = [
        'age' => 'integer',
    ];

    protected $dates = ['birthday'];
}
```

**After**

```php
class Person extends Model
{
    protected $casts = [
        'age' => 'integer',
        'birthday' => 'datetime',
    ];
}
```

### DIFF
--- a/config/sets/laravel57.php
+++ b/config/sets/laravel57.php
@@ -2,6 +2,7 @@
 
 declare(strict_types=1);
 
+use PHPStan\Type\ObjectType;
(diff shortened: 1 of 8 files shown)


TomasVotruba:
Thank you for your contributions :+1:

PR #29, opened 2021-12-03
[Dep] Update requirement to PHP 8.1

samsonasik:


samsonasik:
All checks have passed 🎉 @TomasVotruba I think it is ready.

PR #31, opened 2021-12-08
[Laravel] Set implements AllowEmptyConfigurableRectorInterface on OptionalToNullsafeOperatorRector

samsonasik:
ref https://github.com/rectorphp/rector-src/pull/1422, this Rule is executed, even the `excludeMethods` config not configured.


samsonasik:
I added tag rector-laravel 0.11.16 for it https://github.com/rectorphp/rector-laravel/releases/tag/0.11.16

PR #33, opened 2021-12-21
Add AddGenericReturnTypeToRelationsRector

canvural:
This PR adds a new rector called `AddGenericReturnTypeToRelationsRector` Which adds generic return type docblocks to Laravel relations in models.

Laravel relations are not generic. But Larastan (PHPStan extension for Laravel) makes them generic to be able to do more accurate analysis on relations. So this rector is not really useful for normal Laravel users, but only for Larastan users. I hope you'll still accept it though.

I tried to implement it in a way that it will add `@return` doc tag if
- the method has a relation return type (There are other rectors to add missing return types)
- the method doesn't have any comments at all
- the method has comments but no existing `@return`
- the method only has 1 statement in body and it's `return` keyword. (this is how almost all Laravel relations are implemented)

From my local testing the only error I'm getting is about cognitive complexity. I don't know how I can reduce it more. Any tips for that would be great!

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1,4 +1,4 @@
-# 20 Rules Overview
+# 21 Rules Overview
 
 ## AddArgumentDefaultValueRector
 
@@ -12,14 +12,16 @@ Adds default value for arguments in defined methods.
 use Rector\Laravel\Rector\ClassMethod\AddArgumentDefaultValueRector;
 use Rector\Laravel\ValueObject\AddArgumentDefaultValue;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
-use Symplify\SymfonyPhpConfig\ValueObjectInliner;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
     $services = $containerConfigurator->services();
 
     $services->set(AddArgumentDefaultValueRector::class)
-->configure([                new AddArgumentDefaultValue('SomeClass', 'someMethod', 0, false),
-]);
+        ->configure([
+            AddArgumentDefaultValueRector::ADDED_ARGUMENTS => [
+                new AddArgumentDefaultValue('SomeClass', 'someMethod', 0, false),
+            ],
+        ]);
 };
 ```
 
@@ -31,6 +33,29 @@ return static function (ContainerConfigurator $containerConfigurator): void {
 -    public function someMethod($value)
 +    public function someMethod($value = false)
      {
+     }
+ }
+```
+
+<br>
+
+## AddGenericReturnTypeToRelationsRector
+
+Add generic return type to relations in child of `Illuminate\Database\Eloquent\Model`
+
+- class: [`Rector\Laravel\Rector\ClassMethod\AddGenericReturnTypeToRelationsRector`](../src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php)
+
+```diff
+ use App\Account;
+ use Illuminate\Database\Eloquent\Model;
+ use Illuminate\Database\Eloquent\Relations\HasMany;
+
+ class User extends Model
+ {
++    /** @return HasMany<Account> */
+     public function accounts(): HasMany
+     {
+         return $this->hasMany(Account::class);
      }
  }
 ```
@@ -307,9 +332,9 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $services = $containerConfigurator->services();
 
     $services->set(OptionalToNullsafeOperatorRector::class)
-        ->call('configure', [[
+        ->configure([
             OptionalToNullsafeOperatorRector::EXCLUDE_METHODS => ['present'],
-        ]]);
+        ]);
 };
 ```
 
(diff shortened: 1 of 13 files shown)


docs/rector_rules_overview.md
```diff
@@ -18,8 +18,11 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $services = $containerConfigurator->services();
 
     $services->set(AddArgumentDefaultValueRector::class)
-->configure([                new AddArgumentDefaultValue('SomeClass', 'someMethod', 0, false),
-]);
+        ->call('configure', [[
+            AddArgumentDefaultValueRector::ADDED_ARGUMENTS => ValueObjectInliner::inline([
+                new AddArgumentDefaultValue('SomeClass', 'someMethod', 0, false),
+            ]),
+        ]]);
```
canvural, line 25: This is changed when I ran `composer docs` Let me know if you want me to revert this.

TomasVotruba: It requires this update if composer.json

https://github.com/rectorphp/rector-symfony/commit/041b6a55afbf249daea857459a978dd8308dd72d#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34

canvural: Should that be done in separate PR or in this one?

TomasVotruba: Both options are fine, it's a 1 line fix

src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php
```diff
@@ -0,0 +1,177 @@
...
+
+        if (! $relationMethodCall instanceof MethodCall) {
+            return null;
+        }
+
+        $relatedClass = $this->getRelatedModelClassFromMethodCall($relationMethodCall);
+
+        if ($relatedClass === null) {
+            return null;
+        }
```
samsonasik, line 138: please move some return null code block above to separate private methods based on the code purpose to reduce complexity, eg:

```
private function shouldSkipByObjectType() ...
private function shouldSkipByReturnType() ...
```

canvural: Thank you for the tip! It's reduced now.

src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php
```diff
@@ -0,0 +1,184 @@
...
+        if (! $docInfo instanceof PhpDocInfo) {
+            return null;
+        }
+
+        // Return, if already has return type
+        if ($node->getDocComment() !== null && $docInfo->hasByName('return')) {
+            return null;
+        }
+
+        $returnStatement = $this->betterNodeFinder->findFirstInstanceOf($methodStatements, Return_::class);
```
samsonasik, line 115: You can use `->findFirstInFunctionLikeScoped()` to ensure the return statement is not in the inner function, eg:

```php
        $returnStatement = $this->betterNodeFinder->findFirstInFunctionLikeScoped(
            $node,
            fn (Node $subNode): bool => $subNode instanceof Return_
        );
```

that's check the method stmts, but only if in scoped. By this, the following code will be skipped:

```php
function toBeChecked()
{
   function inner()
   {
         return X;
   }
}
```

src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php
```diff
@@ -0,0 +1,184 @@
...
+        return $modelType->getClassName();
+    }
+
+    private function shouldSkipNode(ClassMethod $node): bool
+    {
+        if ($node->stmts === null) {
+            return true;
+        }
+
+        $classLike = $this->betterNodeFinder->findParentType($node, ClassLike::class);
```
samsonasik, line 177: is it possible on `Interface_` and `Trait_` ?, If not, you can use specific node only if it only possible for 1 node, eg for `Class_` only:

```php
$class = $this->betterNodeFinder->findParentType($node, Node\Stmt\Class_::class);
```

src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php
canvural: Thanks. Changed.

src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php
canvural: Yes, in can happen in trait also. I added a test case for that and update code to be more bulletproof.

TomasVotruba:
@samsonasik Feel free to merge when ready :+1:

samsonasik:
@canvural please run fix-cs command:

```
composer fix-cs
```

to make CI green so I can merge

samsonasik:
Thank you @canvural

canvural:
Thanks for the review and merge @samsonasik

PR #37, opened 2022-03-25
Add new code refactor rule 

hirenkeraliya:
This PR adds a new rule `RedirectBackToBackHelperRector` to refactor the code around redirect back.

Most of the Laravel developers redirect to the previous location with the code like this:

```php
return redirect()->back()->with('...');
```

But there is a Laravel helper called [back](https://laravel.com/docs/9.x/helpers#method-back) that can be called directly and the `redirect()` call is not needed. The above line can be rewritten as:

```php
return back()->with('...');
```

This is my first PR with Rector. Please feel free to share your honest feedback. I am new to AST and related stuff. The Rector book helped me a lot in learning it. Thanks for making  this tool.


samsonasik:
Please run the following commands to fix rector notice:

```
vendor/bin/rector process --ansi
vendor/bin/ecs check --fix --ansi
```

src/Rector/MethodCall/RedirectBackToBackHelperRector.php
```diff
@@ -0,0 +1,137 @@
...
+
+        if (! $parentNode->var instanceof FuncCall) {
+            return null;
+        }
+
+        if ($parentNode->var->getArgs() !== []) {
+            return null;
+        }
+
+        if ($this->getName($parentNode->var->name) !== 'redirect') {
```
samsonasik, line 114: `$parentNode->var->name instanceof Name` may need to be checked first to ensure the call is not `$redirect()->...`

src/Rector/MethodCall/RedirectBackToBackHelperRector.php
```diff
@@ -0,0 +1,137 @@
...
+        $this->removeNode($node);
+
+        $parentNode->var->name = new Name('back');
+
+        return $parentNode;
+    }
+
+    private function updateRedirectStaticCall(StaticCall $node): ?FuncCall
+    {
+        if (! $this->isName($node->class, 'Illuminate\Support\Facades\Redirect')) {
```
samsonasik, line 127: $node->class need to be checked as Name instance, and $node->name need to be checked as Identifier instance

src/Rector/MethodCall/RedirectBackToBackHelperRector.php
samsonasik: Just re-check the `getName()` check on Expr, seems fine for it.

src/Rector/MethodCall/RedirectBackToBackHelperRector.php
samsonasik: Just re-check the `getName()` check on Expr, seems fine for it.

src/Rector/MethodCall/RedirectBackToBackHelperRector.php
samsonasik, line 114: should can use `isName()` check:

```suggestion
        if (! $this->isName($parentNode->var->name, 'redirect')) {
```

samsonasik:
For note, it possibly overlapped/conflict with `ArgumentFuncCallToMethodCallRector` usage on `laravel-static-to-injection.php`

https://github.com/rectorphp/rector-laravel/blob/6bee4284d50e84eb5d027ff266801b2a1c903a33/config/sets/laravel-static-to-injection.php#L207

that the FuncCall changed to method call with dependency injection.

so in `RuleDefinition` description may need to be updated to warn it for its usage with careful.

hirenkeraliya:
> For note, it possibly overlapped/conflict with `ArgumentFuncCallToMethodCallRector` usage on `laravel-static-to-injection.php` config set:
> 
> https://github.com/rectorphp/rector-laravel/blob/6bee4284d50e84eb5d027ff266801b2a1c903a33/config/sets/laravel-static-to-injection.php#L207
> 
> that the FuncCall changed to method call with dependency injection.
> 
> so in `RuleDefinition` description may need to be updated to warn it for its usage with careful.

Hey, I can not get your point. What steps do I need to follow for this pr because of the `laravel-static-to-injection.php` file?
And one more thing I can not get it is `new ArgumentFuncCallToMethodCall('redirect', 'Illuminate\Routing\Redirector', 'back')` why it would be `redirect->back()` instead of the `redirect->to()`?

Please help me with this one. I know it's too much discussion and it's a waste of time for us. sorry, I really apologize for it.

samsonasik:
`laravel-static-to-injection.php` is defined in `LaravelSetList::LARAVEL_STATIC_TO_INJECTION`: 

https://github.com/rectorphp/rector-laravel/blob/6bee4284d50e84eb5d027ff266801b2a1c903a33/src/Set/LaravelSetList.php#L79

so you can use:

```php
use Rector\Laravel\Set\LaravelSetList;
use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $containerConfigurator): void {
    $containerConfigurator->import(LaravelSetList::LARAVEL_STATIC_TO_INJECTION);
};
```

The overlapped may be not 1<->1 apply, but one of rule may not be left unused per-definition.

For usage along with other service, you can do:

```php
use Rector\Laravel\Set\LaravelSetList;
use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $containerConfigurator): void {
    $containerConfigurator->import(LaravelSetList::LARAVEL_STATIC_TO_INJECTION);

    $services = $containerConfigurator->services();
    $services->set(\Rector\Laravel\Rector\MethodCall\RedirectBackToBackHelperRector::class);
};
```

flipped check may be needed:

```php
use Rector\Laravel\Set\LaravelSetList;
use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $containerConfigurator): void {
    $services = $containerConfigurator->services();
    $services->set(\Rector\Laravel\Rector\MethodCall\RedirectBackToBackHelperRector::class);

    $containerConfigurator->import(LaravelSetList::LARAVEL_STATIC_TO_INJECTION);
};
```

to test the overlapped functionality.

hirenkeraliya:
@samsonasik 

I tried this both ways and it's working. But I found an (issue)[https://github.com/rectorphp/rector-laravel/issues/39] that I reported.

samsonasik (approved):

samsonasik:
Thank you @hirenkeraliya

PR #42, opened 2022-04-20
[DX] Update to use Method Call from RectorConfig on rector configs

samsonasik:


tests/Rector/PropertyFetch/OptionalToNullsafeOperatorRector/config/configured_rule.php
```diff
@@ -9,12 +9,10 @@
...
     $rectorConfig->import(__DIR__ . '/../../../../../config/config.php');
 
-    $services = $rectorConfig->services();
-
-    $services->set(OptionalToNullsafeOperatorRector::class)
-        ->call('configure', [
-            [
-                OptionalToNullsafeOperatorRector::EXCLUDE_METHODS => ['present'],
-            ],
-        ]);
```
samsonasik, line 19: @TomasVotruba I tried using `UpgradeRectorConfigRector`, it seems when using old `->call('configure')`, it replace incorrectly, I changed it manually to use direct array values.

rector.php
```diff
@@ -1,20 +1,18 @@
...
 
 return static function (RectorConfig $rectorConfig): void {
-    $parameters = $rectorConfig->parameters();
-
-    $parameters->set(Option::AUTO_IMPORT_NAMES, true);
-    $parameters->set(Option::PATHS, [__DIR__ . '/src', __DIR__ . '/tests']);
-    $parameters->set(Option::SKIP, [
+    $rectorConfig->importNames();
+    $rectorConfig->paths([__DIR__ . '/src', __DIR__ . '/tests']);
+    $rectorConfig->skip([
```
samsonasik, line 15: @TomasVotruba I tried using `UpgradeRectorConfigRector`, it seems `paths()` and `skip()` is not changed by `UpgradeRectorConfigRector` so it changed manually.

config/config.php
```diff
@@ -17,5 +17,5 @@
     $services->load('Rector\\Laravel\\', __DIR__ . '/../src')
         ->exclude([__DIR__ . '/../src/{Rector,ValueObject}']);
 
-    $services->set(RenameClassNonPhpRector::class);
+    $rectorConfig->rule(RenameClassNonPhpRector::class);
```
samsonasik, line 20: @TomasVotruba this needs manual change as well as $services uses at set load 

```

    $services->defaults()
        ->public()
        ->autowire()
        ->autoconfigure();
    $services->load('Rector\\Laravel\\', __DIR__ . '/../src')
        ->exclude([__DIR__ . '/../src/{Rector,ValueObject}']);
```

samsonasik: On rector-symfony usage, it cause error:

```
  rector-symfony git:(main) ✗ vendor/bin/rector process config --dry-run

                                                                                                                        
 [ERROR] Call to undefined method "Rector\Config\RectorConfig::set()" in                                                
         /Users/samsonasik/www/rector-symfony/config/config.php (which is being imported from                           
         "/Users/samsonasik/www/rector-symfony/rector.php").
```

it seems needs to be reverted to use `$services->set()` to avoid same error.

TomasVotruba: :+1:

PR #43, opened 2022-04-20
Revert RectorConfig->rule usage on config/config

samsonasik:
To avoid error like:

```

                                                                                                                        
 [ERROR] Expected an instance of this class or to this class among his parents                                          
         Rector\Core\Contract\Rector\RectorInterface. Got: string in                                                    
         /Users/samsonasik/www/rector-symfony/config/config.php (which is being imported from                           
         "/Users/samsonasik/www/rector-symfony/rector.php").      
```

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -17,5 +17,5 @@ return static function (RectorConfig $rectorConfig): void {
     $services->load('Rector\\Laravel\\', __DIR__ . '/../src')
         ->exclude([__DIR__ . '/../src/{Rector,ValueObject}']);
 
-    $rectorConfig->rule(RenameClassNonPhpRector::class);
+    $services->set(RenameClassNonPhpRector::class);
 };


TomasVotruba:
Thanks 👍

TomasVotruba:
I'm looking into this. When does this error happen? The class should be always present.

samsonasik:
this was happen on running in the rector-symfony itself: `vendor/bin/rector`, possibly due to overlap import and sets call in the rector.php itself ref after re-generate via `UpgradeRectorConfigRector`

 https://github.com/rectorphp/rector-symfony/commit/bc1630b068c4cd0030b26728e50afcb5eb6edc5b#diff-d933a2ad57daa43419b483d3f1ddedaec7aa44933007fead9f9437390f596f4a

https://github.com/rectorphp/rector-symfony/commit/7dd7cdf3850ff4b0a7839fcf24bb4c26138044de#diff-d933a2ad57daa43419b483d3f1ddedaec7aa44933007fead9f9437390f596f4a

TomasVotruba:
I see. Probably old autoload or missing `composer install` run.

It must be valid call. If that happens again somewhere, please point me to it. I'll check it

PR #45, opened 2022-05-14
Add New Refactore Rule - Redirect Route To To Route Helper

hirenkeraliya:
This PR adds a new rule `RedirectRouteToToRouteHelperRector` to refactor the code around the redirect route.

Most of the Laravel developers redirect to the route location with the code like this:

```php
return redirect()->route('home')->with('...');
```

But there is a Laravel helper called [to_route](https://laravel.com/docs/9.x/helpers#method-to-route) that can be called directly and the `redirect()` call is not needed. The above line can be rewritten as:

```php
return to_route('home')->with('...');
```

### DIFF
--- a/src/Rector/MethodCall/RedirectRouteToToRouteHelperRector.php
+++ b/src/Rector/MethodCall/RedirectRouteToToRouteHelperRector.php
@@ -1,0 +1,138 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Laravel\Rector\MethodCall;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Name;
+use Rector\Core\Rector\AbstractRector;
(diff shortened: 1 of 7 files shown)


src/Rector/MethodCall/RedirectRouteToToRouteHelperRector.php
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        if (! $this->isName($parentNode->var->name, 'redirect')) {
+            return null;
+        }
+
+        $this->removeNode($methodCall);
+
+        $parentNode->var->name = new Name('to_route', $methodCall->getArgs());
+        $parentNode->var->args = $methodCall->getArgs();
```
samsonasik, line 121: repetitive `$methodCall->getArgs()` can be moved to variable, eg: `$args = $methodCall->getArgs()` early

hirenkeraliya: My Bad. it's Unnecessary Code.

samsonasik (approved):

TomasVotruba:
Thank you @hirenkeraliya :+1:

PR #46, opened 2022-05-14
Code fixes of the redirect route rector class

hirenkeraliya:
I made too many mistakes on the last PR #45. So, This PR contains the fixes of it.


TomasVotruba:
Thank you for prompt fix :+1:

PR #47, opened 2022-05-14
[Laravel] Add RemoveDumpDataDeadCodeRector

hirenkeraliya:
This PR adds a new rule `RemoveDumpDataDeadCodeRector` to remove the unnecessary dump data code.

We are Humans And Human makes mistake sometimes we miss removing dump data code like `dd` or `dump` function calls.

That's why I added this new rector class to remove code that the developer forgot to remove.

### DIFF
--- a/src/Rector/FuncCall/RemoveDumpDataDeadCodeRector.php
+++ b/src/Rector/FuncCall/RemoveDumpDataDeadCodeRector.php
@@ -6,8 +6,9 @@ namespace Rector\Laravel\Rector\FuncCall;
 
 use PhpParser\Node;
 use PhpParser\Node\Expr\FuncCall;
-use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Stmt\Expression;
 use Rector\Core\Rector\AbstractRector;
+use Rector\NodeTypeResolver\Node\AttributeKey;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 
@@ -68,7 +69,7 @@ CODE_SAMPLE
     }
 
     /**
-     * @param FuncCall|StaticCall $node
+     * @param FuncCall $node
      */
     public function refactor(Node $node): ?Node
     {
@@ -80,7 +81,13 @@ CODE_SAMPLE
             return null;
         }
 
+        $parentNode = $node->getAttribute(AttributeKey::PARENT_NODE);
+        if (! $parentNode instanceof Expression) {
+            return null;
+        }
+
         $this->removeNode($node);
+
         return null;
     }
 }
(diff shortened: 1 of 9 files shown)


src/Rector/FuncCall/RemoveDumpDataDeadCodeRector.php
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * @param FuncCall|StaticCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $node instanceof FuncCall) {
+            return null;
+        }
+
+        if (! $this->isName($node->name, 'dd') && ! $this->isName($node->name, 'dump')) {
```
samsonasik, line 79: `isNames($node, ['dump', 'dd])` can be used

src/Rector/FuncCall/RemoveDumpDataDeadCodeRector.php
```diff
@@ -0,0 +1,86 @@
...
+            ]
+        );
+    }
+
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [FuncCall::class];
```
samsonasik, line 67: It should use Expression instead, so check against Expression expr is FuncCall direct remove FuncCall on If_ cond possibly cause error

samsonasik: Another approach is using FuncCall, but when remove, verify first that parent attribute is Expression stmt

tests/Rector/FuncCall/RemoveDumpDataDeadCodeRector/Fixture/skip_in_if_condition.php.inc
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Rector\Laravel\Tests\Rector\FuncCall\RemoveDumpDataDeadCodeRector\Fixture;
+
+class SkipWithArguments
```
samsonasik, line 5: should be `SkipAsIfCondition`

src/Rector/FuncCall/RemoveDumpDataDeadCodeRector.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * @param FuncCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $node instanceof FuncCall) {
+            return null;
+        }
```
samsonasik, line 78: can be removed, it always a `FuncCall` as defined in `getNodeTypes()`: 

```suggestion
```

samsonasik:
You can fix CI by run:

```
vendor/bin/rector && composer fix-cs
```

samsonasik (approved):

samsonasik:
Could you update PR title to something like : `[Laravel] Add RemoveDumpDataDeadCodeRector`?

hirenkeraliya:
> Could you update the PR title to something like: `[Laravel] Add RemoveDumpDataDeadCodeRector`?

Sorry.

samsonasik:
Thank you @hirenkeraliya

PR #54, opened 2022-07-01
[Lumen] Create Lumen RoutesStringMiddlewareToArrayRector

simivar:
Move from rectorphp/rector-src#2603

### DIFF
--- a/src/Rector/MethodCall/LumenRoutesStringMiddlewareToArrayRector.php
+++ b/src/Rector/MethodCall/LumenRoutesStringMiddlewareToArrayRector.php
@@ -69,7 +69,6 @@ CODE_SAMPLE)]
             return null;
         }
 
-        /** @var Node\Scalar\String_|Node\Expr\Array_ $middlewareValue */
         $middlewareValue = $arrayItem->value;
         if (! $middlewareValue instanceof String_) {
             return null;
--- a/src/Rector/MethodCall/LumenRoutesStringMiddlewareToArrayRector.php
+++ b/src/Rector/MethodCall/LumenRoutesStringMiddlewareToArrayRector.php
@@ -8,6 +8,7 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\Array_;
 use PhpParser\Node\Expr\ArrayItem;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Scalar\String_;
 use Rector\Core\NodeManipulator\ArrayManipulator;
 use Rector\Core\Rector\AbstractRector;
 use Rector\Laravel\NodeAnalyzer\LumenRouteRegisteringMethodAnalyzer;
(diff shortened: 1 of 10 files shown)


TomasVotruba:
I recommend to run tests and coding standard locally. You'll get feedback much faster.

simivar:
@TomasVotruba Any idea why the service container does not pick up class from the new `NodeAnalyzer` directory? From what I see it should pick up everything from `src` excluding `src/Rector`

samsonasik:
You need to import config, like this 

https://github.com/rectorphp/rector-laravel/blob/a6af877719c86988606720636c6414b39bfe217c/tests/Rector/ClassMethod/AddArgumentDefaultValueRector/config/configured_rule.php#L11

samsonasik:
Please run rector and fix cs command:

```
vendor/bin/rector && composer fix-cs
```

src/Rector/MethodCall/LumenRoutesStringMiddlewareToArrayRector.php
```diff
@@ -0,0 +1,150 @@
...
+        if (! $arrayItem instanceof ArrayItem) {
+            return null;
+        }
+
+        /** @var Node\Scalar\String_|Node\Expr\Array_ $middlewareValue */
+        $middlewareValue = $arrayItem->value;
+        if ($middlewareValue instanceof Array_) {
+            return null;
+        }
+
```
samsonasik, line 77: Better to add non String_ node value as well to avoid explode non string, eg: variable.

src/Rector/MethodCall/LumenRoutesStringMiddlewareToArrayRector.php
```diff
@@ -0,0 +1,150 @@
...
+        /** @var string $middlewareString */
+        $middlewareString = $middlewareValue->value;
+        $splitMiddleware = explode('|', $middlewareString);
```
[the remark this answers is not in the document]
simivar: It still can be single middleware as a string without a pipe so I can't return null early, e.g.

```php
$router->get('/', ['uses' => 'HomeController@home', 'middleware' => 'auth']);
```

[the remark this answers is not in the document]
samsonasik: I see

src/Rector/MethodCall/LumenRoutesStringMiddlewareToArrayRector.php
simivar: Done

src/Rector/MethodCall/LumenRoutesStringMiddlewareToArrayRector.php
```diff
@@ -0,0 +1,150 @@
...
+        if (! $array instanceof Array_) {
+            return null;
+        }
+
+        $arrayItem = $this->findItemInArrayByKey($array, 'middleware');
+        if (! $arrayItem instanceof ArrayItem) {
+            return null;
+        }
+
+        /** @var Node\Scalar\String_|Node\Expr\Array_ $middlewareValue */
```
samsonasik, line 72: `@var` no longer needed

samsonasik (approved):

TomasVotruba:
Thank you :clap:

PR #55, opened 2022-07-01
[Lumen] Add LumenRoutesStringActionToUsesArrayRector

simivar:


src/Rector/MethodCall/LumenRoutesStringActionToUsesArrayRector.php
```diff
@@ -0,0 +1,73 @@
...
+            return null;
+        }
+
+        $string = $node->getArgs()[1]
+            ->value;
+        if (! $string instanceof String_) {
+            return null;
+        }
+
+        $node->getArgs()[1]
```
samsonasik, line 68: You can assign to `$node->args[1] = new Arg(` instead of repeating call `->getArgs()`

simivar: I guess the result would be the same and with call to `->getArgs()` this rule has less memory consumption, right?

samsonasik: ->getArgs() is loop to validate Arg node, call twice means double loop.

simivar: Wouldn't then changing to `$node->args[1] = ...` be the best of two worlds?

samsonasik: You can do `$node->args[1]->value = new Array(` of course, not sure if that will make phpstan notice ( I am on mobile now).

simivar: I've change it to `$node->args...`, if you feel like `new Arg(...)` is better feel free to say so and I'll change it. :)

src/Rector/MethodCall/LumenRoutesStringActionToUsesArrayRector.php
```diff
@@ -0,0 +1,71 @@
...
+
+        if (! $this->lumenRouteRegisteringMethodAnalyzer->isLumenRoutingClass($node)) {
+            return null;
+        }
+
+        if (! $this->lumenRouteRegisteringMethodAnalyzer->isRoutesRegisterRoute($node->name)) {
+            return null;
+        }
+
+        $string = $node->args[1]->value;
```
samsonasik, line 62: getArgs() here is fine, just don't call twice

simivar: Done

src/Rector/MethodCall/LumenRoutesStringActionToUsesArrayRector.php
simivar: Done with `new Arg(...)`

simivar:
@samsonasik thank you for the review, should be fine now 🥳

samsonasik (approved):

PR #57, opened 2022-07-18
Update

TomasVotruba:
- bump deps
- misc
- remove AddMockConsoleOutputFalseToConsoleTestsRector, as very narrow rule


samsonasik:
Close re-open to restart build

TomasVotruba:
Could you handle the rest of packages to update composer deps? The naming rule based on type has changed, so the variables has to be updated.

samsonasik:
I will try.

TomasVotruba:
:+1:

PR #70, opened 2022-12-08
Migrate to simplified accessors and mutators in Laravel 9

GeniJaho:
Helloo 😃 

Not sure if this Rule is essential, or even doable, but this is my first attempt at moving old accessors and mutators to the new syntax introduced in Laravel 9.
I've got it working fine with Accessors, tested them in real-life projects, and got good results.
Mutators are a bit more challenging to migrate, and there's some added complexity since we need to put them into the same method now.
The test case 'migrate-mutators.php.inc' is currently failing in a weird way, and the code still needs a lot of refactoring.

Any thoughts, advice, or help is greatly appreciated.
Thank you.


driftingly:
Awesome, I would love to have this.  Not sure when I'll be able to dig it, but I like where this is going.

GeniJaho:
It seems to be working now, both with new accessors and mutators. 
Should this rule be added to a Laravel set list or used as a one-time fix?

PR #71, opened 2022-12-11
AssertStatus to Dedicated Assert method

peterfox:


TomasVotruba:
Test are passing. Could you add the failing test fixture you linked on Twitter?

peterfox:
@TomasVotruba I've created an [example test](tests/Rector/MethodCall/AssertStatusToAssertMethodRector/Fixture/fixture_with_return_by_doctags.php.inc), but the problem doesn't seem to be an issue other than when it's tested on a Laravel app. I might have just set up a demo Laravel app quickly to demonstrate the problem there.

I've also updated the code to fix the PHPStan issues.

Appreciate you looking at this.

TomasVotruba:
Great job :+1: 

The test fixture must fail here, not outside, so we have the bug covered and isolated. Try to figure out the test fixture to fail here.

peterfox:
@TomasVotruba I've tried to understand the problem but with no luck. Ultimately, I created a Laravel project showing that the rule fails.

TomasVotruba:
I see you've put in the lot of effort, so I'm checking the repository to help you out :+1: 


For some reason, the `$response` is seen as `mixed`.
I assume it will be somehow related to Laravel magic autoloading. I've crashed on this hard before and didn't get pass it.

I'd recommend same setup as Larastan has. There will be probably some magic file that you include and PHPStan will see the types.

peterfox:
Thank you @TomasVotruba. I really appreciate the help. It's odd because, as far as I can tell, it should be picking up the return tag to know that it's a `TestResponse` class, but you're right, it's detecting it as mixed in the real-world example. It seems to do this with Larastan installed as well. It might be something Larastan itself doesn't pick up. I'm sure I can find some way to resolve this, if not  Laravel 10 will add return types I believe, so it might just be something for the future if I can't right now.

peterfox:
@TomasVotruba, just a heads up, [I managed to fix the problem with a tweak to Larastan, but debugging PHPStan, it seems there might just be a bug with Rector](https://github.com/nunomaduro/larastan/pull/1481)?

To summarise. When running PHPStan, it debugs the correct type `TestResponse`, but when running with Rector, it finds it as a mixed type.

Was curious to know if you had any thoughts on how that might of happened and how you might debug that within Rector.

TomasVotruba:
That's probably due to the type extensions, possibly loaded magically via phpstan/extension-installer. 
They must be included in `phpstan-for-rector.neon` too.

peterfox:
It isn't installed in the demo project.

driftingly:
Thanks for your work on this @peterfox.
I took a crack at it and wasn't able to figure out the issue. PHPStan knows the type with/without Larastan.  I don't understand why Rector would have an issue there.

driftingly:
I tested this again in the demo app you provided @peterfox, it works with and without your larastan fork.  I'm going to merge this in now.  Thanks 🙌

PR #75, opened 2023-01-06
Support updating an existing return type with generics

rene-bos:
Only supported if the existing PHPDoc return type is the same as the native return type. This way we ensure that we don't change types. 

Not updating of another type was already covered with a test. I added a test for the case when we can update the PHPDoc with only the generics.


rene-bos:
Forgot to run PHPStan local before the last commit, will fix tomorrow.

ratenfanger:
@driftingly This rule is ready to get reviewed (and hopefully to get merged!). I'm working on a further improvement of this rule which is build on top of this PR. I will prepare the PR if this one merged. That PR will fix the generics for BelongsTo and MorphTo, which both need a TChildModel specified.

rene-bos:
@driftingly This rule is ready to get reviewed (and hopefully to get merged!). I'm working on a further improvement of this rule which is build on top of this PR. I will prepare the PR if this one merged. That PR will fix the generics for BelongsTo and MorphTo, which both need a TChildModel specified.

driftingly:
Thanks @rene-bos!

PR #76, opened 2023-01-14
Use full names for Facade aliases

GeniJaho:
This rector will replace the facade aliases like `use DB;` with the full name of the facade, as defined in `config/app.php -> 'aliases'` in default Laravel installations.

Not entirely sure if that's a good thing, but the editor offers autocompletion when using the full name, and actually recognizes the facade.


GeniJaho:
I think there's some issue with this last step `vendor/bin/easy-ci check-active-class --ansi`. I don't think it's related to the PR itself.

driftingly:
Hi @GeniJaho,

Would you be able to add some tests?
Would this cause false positives for classes in the same directory?
If I had two files, `App\Helpers\Action` and I imported `App\Helpers\Event` as `use Event` would this get converted to `use Illuminate\Support\Facades\Event`?

GeniJaho:
The ways you would import the Event helper, in that case, would be with the full namespace:
`use App\Helpers\Event;` or `use App\Helpers\Event as Event;`.
In either case, the rule does not touch that piece of code.
It would only modify the code if it found an instance of exactly `use Event;`. Not sure how often this can occur though.

This PR no longer introduces a new rule actually and the rule can be applied by anyone as a one-time fix. That said, it should probably be made skippable from the 'laravel-code-quality' set and maybe also configurable. I don't know how, yet.
Since the rule already exists, maybe this PR is totally redundant. I'm OK with closing it if so 😃.

driftingly:
Thanks @GeniJaho 🙌

PR #77, opened 2023-01-23
Add ReplaceFakerInstanceWithHelperRector

GeniJaho:


driftingly:
@GeniJaho can you expand on the TODO there "skip $this->faker->randomEnum(...)."?

GeniJaho:
> @GeniJaho can you expand on the TODO there "skip $this->faker->randomEnum(...)."?

I'm using this package https://github.com/spatie/laravel-enum#faker-provider in one of my projects, and I'm using `$this->faker->randomEnum(...)` from its faker provider. However, it does not work if we swap it with `fake()->randomEnum(...)`.

driftingly:
Thanks @GeniJaho 🙌

PR #84, opened 2023-02-07
Fix code-style using `fix-cs` command

rene-bos:
This way contributors don't get errors when using `composer run check-cs` locally.


driftingly:
Thanks!  And thanks for getting us back here:
Screenshot 2023-02-07 at 8 00 03 AM

rene-bos:
All green! 🎉

PR #85, opened 2023-02-07
Don't run Rector on forks

rene-bos:
Running Rector on forks gives this the error message below:

```
Invalid workflow file: .github/workflows/rector.yaml#L14
The workflow is not valid. .github/workflows/rector.yaml (Line: 14, Col: 27): Invalid secret, ACCESS_TOKEN is not defined in the referenced workflow.
```

Futhermore the contributer gets an (email) notification about a failed PR check: ```[driftingly/rector-laravel] PR run failed at startup: Rector - Fix `check-active-classes` and re-enable code_analysis workflow (dd5284a)```. This is very anoying. Other Rector extensions just exclude Rector for forks, thats why I just [grabbed the one from rector-doctrine](https://github.com/rectorphp/rector-doctrine/blob/main/.github/workflows/code_analysis.yaml). All rector-* packages now removed rector.yaml in fafor of the Rector job code_analisis.yaml, so I made that change here also.


rene-bos:
This is the last one (for now 😉).

driftingly:
Looks good!  Thanks @rene-bos

PR #82, opened 2023-02-07
Add TChildModel for relations with childs in `AddGenericReturnTypeToRelationsRector`

rene-bos:

PR #86, opened 2023-02-07
Fix Rector job on PR to main repository

rene-bos:
Run only on PR's from branches on main repository (from core contributors), not on the forks or PR's from forks.

### DIFF
--- a/.github/workflows/code_analysis.yaml
+++ b/.github/workflows/code_analysis.yaml
@@ -11,8 +11,8 @@ jobs:
         uses: rectorphp/reusable-workflows/.github/workflows/code_analysis.yaml@main
 
     rector:
-        # run only on main repository, not on the forks without access
-        if: github.repository == 'driftingly/rector-laravel'
+        # run only on PR's from branches on main repository (from core contributors), not on the forks or PR's from forks.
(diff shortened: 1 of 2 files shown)


driftingly:
I might have merged this too early. Sorry!

rene-bos:
No problem! It should have created it as draft if I did not wanted it to get merged. Its  fine to be merged I think. If TomasVotruba agrees to move the check to rector.yaml we can just remove the if-statement.

For now we can keep the change here, as it will avoid errors like we had on #82.

driftingly:
Should we revert https://github.com/driftingly/rector-laravel/pull/86/commits/3f57062b4ea94d83834e2024355e205bc31d9f1e

rene-bos:
Yeah that's a good one. When we don't fix it it will pop up in a totally unrelated pull request some time 😉 Because of some strange hickup the change has already been committed to #82 ([[ci-review] Rector Rectify](https://github.com/driftingly/rector-laravel/pull/82/commits/75aa0941ddd06ff4c75bff5f2ce06282399b3e91)). 

Do you have time to check #82 and maybe create a release afterwards? At my company (Visma) we would love to have that change 😄. Avoids pinning to our fork and the associated maintenance of it. Sorry for making your day busy with all that PR's 😉

Issue #88, opened 2023-02-14
Support for Laravel 10?


GeniJaho:
Is there any work being done behind the scenes to support the migration to the current release of Laravel 10?

If not, I'd be glad to open a PR for it. Thank you 😁.

driftingly:
That would be amazing @GeniJaho 🙌

driftingly:
Closed by #89

PR #89, opened 2023-02-17
Add Laravel 10 upgrade rules

GeniJaho:
Issue https://github.com/driftingly/rector-laravel/issues/88
Not much to upgrade for Laravel 10, aside from a few renames.

https://laravel.com/docs/10.x/upgrade#database-expressions
Not sure if we should introduce a rule for this. What do we put for the query grammar parameter?

https://laravel.com/docs/10.x/upgrade#redirect-home
If a rule is needed for this, would we redirect to '/'?

https://laravel.com/docs/10.x/upgrade#service-mocking
This seems difficult to be done properly, taking into account the methods are different on the facades.

https://github.com/laravel/framework/pull/41136/files
`Notification::assertTimesSent()` to `Notification::assertSentTimes()` with flipped arguments.
This could be done with a custom rule, but there already exists a `MethodCallRename` rule that renames methods and has all the boilerplate. We need, however, to swap the arguments in an atomic action together with renaming.
Not sure what's the best way to achieve this. Maybe introducing a new `MethodCallRenameWithArgumentSwapping` rule in the core Rector repo, or just this repo?

### DIFF
--- a/.github/workflows/downgraded_release.yaml
+++ b/.github/workflows/downgraded_release.yaml
@@ -1,0 +1,52 @@
+name: Downgraded Release
+
+# https://tomasvotruba.com/blog/how-to-release-php-81-and-72-package-in-the-same-repository/
+# https://github.com/TomasVotruba/cognitive-complexity/blob/main/.github/workflows/downgraded_release.yaml
+# https://github.com/symplify/config-transformer/blob/main/.github/workflows/downgraded_release.yaml
+
+on:
+    push:
+        tags:
+            - '*'
+
+jobs:
+    downgrade_release:
+        runs-on: ubuntu-latest
+
+        steps:
+            -
+                uses: "actions/checkout@v3"
+
+            -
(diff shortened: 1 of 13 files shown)


src/Set/LaravelSetList.php
```diff
@@ -78,6 +78,11 @@ final class LaravelSetList implements SetListInterface
      */
     final public const LARAVEL_90 = __DIR__ . '/../../config/sets/laravel90.php';
 
+    /**
+     * @var string
+     */
+    final public const LARAVEL_10 = __DIR__ . '/../../config/sets/laravel10.php';
```
alexey-m-ukolov, line 84: Shouldn't it be `LARAVEL_100`?

driftingly: Good catch @alexey-m-ukolov 🙌 
I wouldn't have caught that.

driftingly:
Thanks @GeniJaho 🙌

PR #101, opened 2023-04-26
Route callable improvement

egorbwork:
Allows refactoring of routes using the array with `uses` as a single key and arrays that outside of the `name` key are using `middleware` one.


driftingly:
@egorbwork are you able to get those workflows passing?

egorbwork:
@driftingly  Will try as fast as possible.

egorbwork:
@driftingly 
unit tests are fine now:
image

PHPstan I hope too.

driftingly:
Thanks @egorbwork 🙌

Issue #100, opened 2023-04-10
Laravel Cashier upgrade rules


GeniJaho:
Can we include rules to upgrade Cashier to later versions like 13 and 14 in this repository? At least to handle the boring method renames and whatever work that can be automated.

Since it's a first-party Laravel package it would make sense to include it in this repo, but while adding new classes that are Cashier-specific, the Rector rules become bloated and hard to navigate.

A solution could be to prepend the rules with the Cashier prefix, for example, `CashierStripeOptionsToStripeRector`. 

Or would a new repository be a better idea?

driftingly:
Closed by #103

PR #103, opened 2023-04-29
Add Laravel Cashier 13 & 14 upgrade rules

GeniJaho:
Relates to https://github.com/driftingly/rector-laravel/issues/100

This PR introduces initial support for upgrading Laravel Cashier to version 13 and 14.

It only handles the method renames, since [other upgrades](https://github.com/laravel/cashier-stripe/blob/master/UPGRADE.md#upgrading-to-130-from-12x) are out of the scope (and capability) of Rector.


GeniJaho:
Should I include the `LaravelCashierSetList` class somewhere? The CI fails since the class is currently unused in the project.

driftingly:
@GeniJaho What do you think of having `config/sets/packages/cashier-13.php`?
If other package-specific sets are developed we could group them there?

GeniJaho:
> 

Yes, that would be a lot better. Any idea to make the failing check to pass? The CashierSetList class is not used anywhere.

driftingly:
@GeniJaho sorry, that was my suggestion for the set list.  Having something like what you did for Laravel 10:
https://github.com/driftingly/rector-laravel/blob/main/config/sets/level/up-to-laravel-100.php

driftingly:
@GeniJaho is this ready?  
If it's ready to go can you merge in main and update the docs?
Thanks!

GeniJaho:
@driftingly this is ready now, I updated the docs and main was merged. Not sure why that test failed before the merge though, I couldn't make it fail locally.

PR #105, opened 2023-05-11
Adds Gone, InternalServerError & ServiceUnavailable statuses to AssertStatusToAssertMethodRector

peterfox:
# Changes

* Adapted the `AssertStatusToAssertMethodRector` to handle 410, 500 and 503 status codes.
* Updates the test to reflect the changes.
* Updates the test to use a static data provider method.

# Why
Updates the `AssertStatusToAssertMethodRector` with the [new HTTP assertion methods introduced in 10.9](https://laravel-news.com/laravel-10-9-0).

There might be some issues here with a requirement to configure which version of Laravel you're using otherwise you might end up with broken method calls because the method doesn't exist. I'm not sure of the best way to apply this. A configuration to specify which HTTP statuses to refactor.

There's also a minor fix to make the test for this use a static data provider.


driftingly:
Thanks @peterfox 🙌

Issue #78, opened 2023-01-25
`RedirectRouteToToRouteHelperRector` error on calls without `->with`


driftingly:
The following fixture causes an error when run with the `RedirectRouteToToRouteHelperRector` rule.

Seems to be caused by `$this->removeNode($methodCall);`

```php
<?php

namespace RectorLaravel\Tests\Rector\MethodCall\RedirectRouteToToRouteHelperRector\Fixture;

use Illuminate\Support\Facades\Redirect;

class Fixture
{
    public function store()
    {
        return redirect()->route('home');
    }

    public function update()
    {
        return Redirect::route('home');
    }
}

?>
-----
<?php

namespace RectorLaravel\Tests\Rector\MethodCall\RedirectRouteToToRouteHelperRector\Fixture;

use Illuminate\Support\Facades\Redirect;

class Fixture
{
    public function store()
    {
        return to_route('home');
    }

    public function update()
    {
        return to_route('home');
    }
}

?>
```

driftingly:
@hirenkeraliya Do you have any suggestions on this?  I know it's been a while since you worked on this.

hirenkeraliya:
@driftingly Sure I will check and update you regarding this.

ljanecek:
I encountered the same error when I have `redirect()->back();` in the controller without using -`>with`. Unfortunately, I don't yet have sufficient knowledge of the code to be able to create a pull request. 😇

`[ERROR] Could not process "/var/www/app/Http/Controllers/CountryCurrencyController.php" file, due to:                  
         "System error: "leaveNode() returned invalid value of type integer"                                            
         Run Rector with "--debug" option and post the report here: https://github.com/rectorphp/rector/issues/new". On line: 150`

hirenkeraliya:
@driftingly I am trying to fix this issue for the last couple of days. I tried too many things and the issue is the same. Can you please help me out with this?

The issue occurs when we try to remove the last method call of the chaining methods.

driftingly:
Closed by #106

PR #106, opened 2023-05-17
Fix Issue #78 - Error in Redirect::route() Method

hirenkeraliya:
This pull request resolves issue #78. The issue was related to an error occurring when using the `redirect()->route()` method. The bug has been identified and fixed.

**Changes:**

- Analyzed the root cause of the error and determined that it was due to incorrect parameter handling in the `redirect()->back()` method.
- Refactored Some old files code

**Testing:**

Added new unit tests specifically targeting the scenario described in issue #78 to verify the fix.
Ran the full test suite locally to ensure that all existing tests pass successfully.
Manually tested the affected functionality, verifying that the error no longer occurs.


driftingly:
Thanks @hirenkeraliya 🙌

hirenkeraliya:
@driftingly Update me if you want to add any other refactorings or any other changes.

PR #108, opened 2023-05-27
Move `ArgumentFuncCallToMethodCallRector` to package

driftingly:
This PR moves the `ArgumentFuncCallToMethodCallRector` class from Rector to Rector Laravel. 

See https://github.com/rectorphp/rector-src/pull/3774#issuecomment-1539905216

### DIFF
--- a/src/Rector/FuncCall/ArgumentFuncCallToMethodCallRector.php
+++ b/src/Rector/FuncCall/ArgumentFuncCallToMethodCallRector.php
@@ -1,0 +1,254 @@
+<?php
+
+declare(strict_types=1);
+
+namespace RectorLaravel\Rector\FuncCall;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\PropertyFetch;
+use PhpParser\Node\Stmt\Class_;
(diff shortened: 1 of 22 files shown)


src/Rector/FuncCall/ArgumentFuncCallToMethodCallRector.php
```diff
@@ -0,0 +1,254 @@
...
+     * @param FuncCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if ($this->shouldSkipFuncCall($node)) {
+            return null;
+        }
+
+        /** @var Class_ $classLike */
+        $classLike = $this->betterNodeFinder->findParentType($node, Class_::class);
```
samsonasik, line 105: instead of find parent type, it is better to use `Class_` node, and traverser its methods, something like this:

```php
        $hasChanged = false;
        $this->traverseNodesWithCallable($node->getMethods(), function (Node $node) use (
            &$hasChanged
        ) {
                if ($node instanceof  FuncCall) {
                       $result = $this->refactorFuncCallToMethodCall($node);
                       if ($result instanceof MethodCall) {
                              $hasChanged = true;
                              return $result;
                       }

                       return null;
                }

                return null;
        });

        if ($hasChanged) {
              return $node;
        }

        return null;
```

PR #109, opened 2023-06-01
Filter out internal PHP functions from set

### LINKED ISSUE #107: \str_contains should not be replaced by Str::contains when on PHP 8+
FuncCallToStaticCallRector is configured to replace str_contains with Str::contains, but this is also is done on PHP 8, where the native str_contains is used.
Could we add a (target) version check at https://github.com/driftingly/rector-laravel/blob/02c1eb7fe7eb8cacd216cefe77d928237f1f343f/config/sets/laravel-array-str-functions-to-static-call.php#L46, so it will be skipped when the native function is used?

Can you tell me in what way you would like me to create a PR if you're open to such a PR?

driftingly:
This PR filters out internal PHP functions from the `laravel-array-str-functions-to-static-call` set.

Closes #107


johnbacon:
Hi @driftingly, could we make filtering out internal PHP functions optional?

There are often differences between the helpers and related PHP functions. For example, [the inspiration for this PR](https://github.com/driftingly/rector-laravel/issues/107) (str_contains) doesn't accept an array of strings. We'd like to standardize the use of static helpers when possible.

Issue #115, opened 2023-06-17
Cannot autowire service "RectorLaravel\Rector\Assign\CallOnAppArrayAccessToStandaloneAssignRector": argument  "$nodesToAddCollector" of method "__construct()" has type "Rector\PostRector\Collector\NodesToAddCollector" but this class was not found.


khalyomede:
I just upgraded to rector `0.17.1` and I have this error when running `rector process --dry-run`

```bash
rector process --dry-run

                                                                                                                        
 [ERROR] Cannot autowire service "RectorLaravel\Rector\Assign\CallOnAppArrayAccessToStandaloneAssignRector": argument   
         "$nodesToAddCollector" of method "__construct()" has type "Rector\PostRector\Collector\NodesToAddCollector" but
         this class was not found.                                                                                      
                                                                                                                        

Script rector process --dry-run handling the scan event returned with error code 1
```

Here is my `rector.php` file

```php
declare(strict_types=1);

use Rector\Config\RectorConfig;
use Rector\DeadCode\Rector\Node\RemoveNonExistingVarAnnotationRector;
use Rector\Php70\Rector\StaticCall\StaticCallOnNonStaticToInstanceCallRector;
use Rector\Php80\Rector\FuncCall\ClassOnObjectRector;
use Rector\Php81\Rector\Array_\FirstClassCallableRector;
use Rector\Php81\Rector\MethodCall\SpatieEnumMethodCallToEnumConstRector;
use Rector\Php82\Rector\Class_\ReadOnlyClassRector;
use Rector\Privatization\Rector\Class_\ChangeReadOnlyVariableWithDefaultValueToConstantRector;
use Rector\Set\ValueObject\LevelSetList;
use Rector\Set\ValueObject\SetList;
use RectorLaravel\Set\LaravelSetList;

return static function (RectorConfig $rectorConfig): void {
    $rectorConfig->paths([
        __DIR__ . "/app",
        __DIR__ . "/config",
        /**
         * @todo Try to uncomment and see if it still gets stuck in infinite loop (since 0.15.10)
         */
        // __DIR__ . "/database",
        __DIR__ . "/lang",
        __DIR__ . "/resources/views",
        __DIR__ . "/routes",
        __DIR__ . "/tests",
    ]);

    $rectorConfig->skip([
        ClassOnObjectRector::class,
        StaticCallOnNonStaticToInstanceCallRector::class,
        FirstClassCallableRector::class,
[...]

PR #117, opened 2023-06-26
Support for Rector 0.17.1

GeniJaho:
Issue https://github.com/driftingly/rector-laravel/issues/115.


u01jmg3:
@driftingly: any timeline on when this could get merged so we can use the latest version of Rector?

driftingly:
@u01jmg3 this was merged, but it looks like there are more issues with `0.17.3`

See #120.

u01jmg3:
Ah, ever moving 🎯

Issue #121, opened 2023-07-13
Incompatible with Rector v0.17


freshleafmedia:
[Rector v0.17.0](https://github.com/rectorphp/rector/releases/tag/0.17.0) and above aren't compatible. It appears to be due to things like the `findParentType` method and `nodesToAddCollector` property being removed.

I'm using rector-laravel v0.21.0

gisostallenberg:
`composer.json` should contain the required rector version.

I run driftingly/rector-laravel v0.21.0 which works fine on rector 0.17.2 but not on 0.17.3, because in that version BetterNodeFinder::findParentType was removed, see https://github.com/rectorphp/rector/releases/tag/0.17.3

According to [SemVer](https://semver.org/#spec-item-4) rector can do breaking changes, but I think it should have a new minor version (0.18.0).

driftingly:
@gisostallenberg @freshleafmedia Just released the updates by @GeniJaho, which should resolve this issue.  Let me know.

gisostallenberg:
@driftingly you mean the changes in https://github.com/driftingly/rector-laravel/pull/117?

driftingly:
Yeah. I tried on a fresh install and latest Rector and it ran. Noticed there are some failing test though with everything merged in.

driftingly:
And it looks like we are green again 🙌

Issue #120, opened 2023-07-10
Support for Rector 0.17.3


driftingly:
It looks like Rector is moving away from BetterNodeFinder.
Lots of deprecated methods: https://github.com/rectorphp/rector/issues/7947

We'll need to update calls to `$this->betterNodeFinder->*`.

GeniJaho:
Do you have any suggestions about finding parents of classes? With BetterNodeFinder it was a one-liner but now we have to rely on reflection, which adds a few extra lines and cases to think about. I'm working on updating the rules dependent on BetterNodeFinder but not sure if there's a better way.

Also not sure if there are any performance penalties if we use AbstractScopeAwareRector instead of AbstractRector 👀.

driftingly:
Thanks for all your work on this @GeniJaho 🙌

PR #122, opened 2023-07-15
Core rector support - attempt #2

GeniJaho:
https://github.com/driftingly/rector-laravel/issues/120
https://github.com/driftingly/rector-laravel/issues/121
This PR fixes most of the errors preventing the upgrade up to Rector 0.17.6, except for RequestStaticValidateToInjectRector, which is turning out to be a different beast entirely. I'll try to fix it in a few days, maybe in another PR.


GeniJaho:
@driftingly The last commit is really a 'downgrade' of the RequestStaticValidateToInjectRector rule. It now only works when we're looking at assignment statements (`$validated = request()->validate();`) or direct statements (`request()->validate();`).
The reason for that is that now we have to iterate the statements of the ClassMethods and find the `request()` or `Request::` calls 'by hand' (the replace() method I've introduced is an abomination 😁).

I couldn't find a better way, but this solution worked in the few projects it tested it.

driftingly:
Thanks for your work on this @GeniJaho 
I tried to figure out the required changes by what they changed in the rector codebase but couldn't.  I hope they release a 1.0 soon and we can work on improving these rules.  

Looks like some work was done on
https://github.com/cakephp/upgrade/pull/235
https://github.com/cakephp/upgrade/pull/236/files

Might be able to use some of that for inspiration/direction.

GeniJaho:
@driftingly Thank you so much for the tip, `$this->traverseNodesWithCallable()` did the trick. The `RequestStaticValidateToInjectRector` rule now works exactly the same as before, with minimal changes to the class.
This branch now works with the latest release of Rector 0.17.7.

driftingly:
@GeniJaho Do those `skip_*` tests need to be renamed?

GeniJaho:
> @GeniJaho Do those `skip_*` tests need to be renamed?

No, they were actually there before and are relevant to that Rule, the Rule should not affect usages in traits or extends.
They appear as new files, but they're just modified.

Fixtures for unchanged functionality don't need to define two identical parts, like this:
```php
<?php
$some = 'test';
?>
---
<?php
$some = 'test';
?>
```
If we don't put a separator at all, that tells Rector that this piece of code shouldn't be changed.
```php
<?php
$some = 'test';
?>
```
So I just simplified those two fixtures, they're appearing as new files, however.

PR #127, opened 2023-08-17
Upgrade service type setting in config.php

### LINKED ISSUE #125: Rector 0.17.14 requires syntax change
**PHP Fatal error:  The services() method is deprecated. Use $rectorConfig->singleton(ServiceType::class) instead**

https://github.com/driftingly/rector-laravel/blob/eb9de6d98bbf899db1eba0a2d21640422d5b9520/config/config.php#L10

szepeviktor:
Service type may be one of these
https://github.com/rectorphp/rector-src/blob/5363a676265cbe868f4ecc6bde8c6a99b213d028/src/DependencyInjection/LazyContainerFactory.php#L328-L335

`RenameClassNonPhpRector` [does not exist](https://github.com/rectorphp/rector-src/commit/1659ca292848fc1722b4b084d9fd1625b21e69c9) anymore.

Fixes #125

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -7,7 +7,7 @@ use Rector\Config\RectorConfig;
 use Rector\Core\NonPhpFile\Rector\RenameClassNonPhpRector;
 
 return static function (RectorConfig $rectorConfig): void {
-    $services = $rectorConfig->services();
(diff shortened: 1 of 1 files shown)


driftingly:
Thanks for jumping on this @szepeviktor. 
This was done in #126

PR #130, opened 2023-08-22
Remove reverted Laravel 10 rule (getBaseQuery -> toBase)

### LINKED ISSUE #129: `getBaseQuery` -> `toBase` was reverted in Laravel 10
The deprecated `getBaseQuery` was initially removed, but then it was [added back](https://github.com/laravel/framework/pull/41923) because it caused a regression.

I think this should be removed from the Laravel 10 set.

joelvh:
The deprecated `getBaseQuery` was initially removed, but then it was [added back](https://github.com/laravel/framework/pull/41923) because it caused a regression.

Closes #129

### DIFF
--- a/config/sets/laravel100.php
+++ b/config/sets/laravel100.php
@@ -31,8 +31,6 @@ return static function (RectorConfig $rectorConfig): void {
 
     $rectorConfig
         ->ruleWithConfiguration(RenameMethodRector::class, [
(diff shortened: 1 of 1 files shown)


joelvh:
FYI the Laravel [upgrade docs](https://github.com/laravel/docs/pull/8970) have been updated to _remove_ this change from the list

PR #131, opened 2023-08-22
Correct paths

u01jmg3:
RE: PR #103

Looks like these paths weren't updated when the files were moved based on PR feedback


driftingly:
Thanks @u01jmg3 🙌

PR #148, opened 2023-10-12
Allow configuring EloquentOrderByToLatestOrOldestRector

### LINKED ISSUE #147: Allow configuration with `EloquentOrderByToLatestOrOldestRector` rule
By default, [EloquentOrderByToLatestOrOldestRector](https://github.com/driftingly/rector-laravel/blob/3054e6e8a11baf41a17e640804dd5e064b674a9d/src/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector.php) modifies all `orderBy` and `orderByDesc` methods to use `latest()` or `oldest()` instead.

Our team feels we'd like to do this only regarding timestamps, as `oldest('name')` or similar could be confusing.

As such, some configuration allowing for explicit column names you'd like to change to `oldest()` or `latest()` is desired.

----

This suggestion was also briefly mentioned in [the relevant PR](https://github.com/driftingly/rector-laravel/pull/142#issuecomment-1752137821):
> These changes apply to all columns in our project, not just date/datetime columns. Is that intended behavior? If so, I'm wondering if rector-laravel is open to configuration, so we can specify which columns should and shouldn't have these rules applied.

johnbacon:
Resolves #147.

- Allows an `ALLOWED_PATTERNS` array of strings to apply this rule to
- Adds additional test items for variable handling

#### Notes

- The way this is implemented, it will always change `created_at` columns (as it did before)
- Because static analysis cannot get the value of a variable, we instead allow for specifying variable names

#### Tested

- Current test scenario
- Scenario where no rule configuration is present

### DIFF
--- a/src/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector.php
+++ b/src/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector.php
@@ -174,6 +174,7 @@ CODE_SAMPLE
     {
         $allowedPatterns = $configuration[self::ALLOWED_PATTERNS] ?? [];
         Assert::isArray($allowedPatterns);
+        Assert::allString($allowedPatterns);
 
         $this->allowedPatterns = $allowedPatterns;
     }
--- a/src/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector.php
+++ b/src/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector.php
@@ -10,6 +10,7 @@ use PHPStan\Type\ObjectType;
 use Rector\Core\Contract\Rector\ConfigurableRectorInterface;
 use Rector\Core\Rector\AbstractRector;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\ConfiguredCodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 use Webmozart\Assert\Assert;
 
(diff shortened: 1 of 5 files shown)


driftingly:
We use `composer docs` to update the docs using `symplify/rule-doc-generator`.
Can you ensure your changes are compatible?

johnbacon:
@driftingly That should be done now! Thanks for the heads up, it wasn't compatible without changes.

src/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector.php
```diff
@@ -107,4 +165,16 @@ private function convertOrderByToLatest(MethodCall $methodCall): MethodCall
...
     }
+
+
+    /**
+     * @param mixed[] $configuration
+     */
+    public function configure(array $configuration): void
+    {
+        $allowedPatterns = $configuration[self::ALLOWED_PATTERNS] ?? [];
+        Assert::isArray($allowedPatterns);
```
peterfox, line 176: I think there should also be an assertion here to check all values in the array are strings.

tests/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector/Fixture/fixture.php.inc
```diff
@@ -11,7 +11,9 @@ $query->orderBy('created_at');
 $query->orderBy('created_at', 'desc');
 $query->orderBy('submitted_at');
 $query->orderByDesc('submitted_at');
-$query->orderBy($column);
+$query->orderBy($allowed_variable_name);
+$query->orderBy($unallowed_variable_name);
+$query->orderBy('unallowed_column_name');
```
peterfox, line 16: Would probably be better to exist in their own file like [skip_if_not_eloquent_builder.php.inc](https://github.com/driftingly/rector-laravel/blob/main/tests/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector/Fixture/skip_if_not_eloquent_builder.php.inc)

```php
// skip_if_column_not_allowed.php.inc
<?php

namespace RectorLaravel\Tests\Rector\Cast\DatabaseExpressionCastsToMethodCall\Fixture;

$query->orderBy($unallowed_variable_name);
$query->orderBy('unallowed_column_name');

?>
```

peterfox (changes requested):
a couple things I'd consider changing

src/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector.php
johnbacon: Good call, done!

tests/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector/Fixture/fixture.php.inc
johnbacon: I can do this, but I thought these would be unallowed not because they weren't an instance of Builder but because they weren't desired to be auto-renamed. Does it still make sense to move to a different file? I'm not familiar with best-practices in this regard.

driftingly: I think @peterfox is saying to create a file called `skip_if_column_not_allowed.php.inc` or `skip_if_column_not_configured.php.inc` similar to how he did with with the [EloquentOrderByToLatestOrOldestRector](https://github.com/driftingly/rector-laravel/tree/main/tests/Rector/MethodCall/EloquentOrderByToLatestOrOldestRector) fixtures.

PR #156, opened 2023-12-15
Update code style

driftingly:


driftingly:
@GeniJaho You don't have to review this whole PR. Most of it is just updates to styles and some actions.  Switch over to use Duster which is a tool I maintain for code style.  It's similar to Easy Coding Standard but adds Pint and TLint. 
Do you have any thoughts this move?

GeniJaho (approved):
Hey that's fantastic, and PHP 8.3 as well 💪

PR #162, opened 2024-01-01
[Laravel100] Adds ReplaceExpectsMethodsInTests rule

peterfox:
# Changes

* Adds the new rule to replace `expectsJobs` and `expectsEvents`.
* Adds tests
* Updates docs
* Updates the Laravel100 config to include the new rule

# Why

It's mentioned in the [Laravel Upgrade guides](https://laravel.com/docs/10.x/upgrade#service-mocking) and this rule is quite useful in that it will insert the assertions at the end of the tests' method statements as the best way of handling this change via Rector.

# Notes

I can't make this work for `expectsNotifications` as notification assertions require the notifiable to be checked and as there is no way to know the context of that from a test, only the jobs and events can be handled properly via Rector. The closest guess would be to use `assertSentTimes` and put 1 as a default which if that's the case it would be best done with an additional rule.


driftingly:
I've pulled in the latest changes to `main` here.  Updates to coding style and GitHub actions.

PR #172, opened 2024-01-25
[Fix] EloquentMagicMethodToQueryBuilderRector provides incorrect doc sample

peterfox:
# Changes

* Corrects the sample docs of the EloquentMagicMethodToQueryBuilderRector rule
* Docs updated.

# Why

Without this fix the `composer docs` script fails to generate the new rules documentation for the repo

# Notes

I'm not sure why the command is now removing lots of examples. Possible that they've been added manually?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "phpstan/phpstan": "^1.8.2",
         "symplify/phpstan-rules": "^12.4",
         "symplify/phpstan-extensions": "^11.0",
-        "symplify/rule-doc-generator": "^11.0",
+        "symplify/rule-doc-generator": "^12.1",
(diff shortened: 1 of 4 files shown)


driftingly:
Thanks @peterfox, looks like updating `symplify/rule-doc-generator` to the latest mostly fixes the issue:

```diff
-        "symplify/rule-doc-generator": "^11.0",
+        "symplify/rule-doc-generator": "^12.1.1",
```

There are still a few blocks removed, but those appear to be where the docs generator would insert an example configuration.  Looks like it doesn't do that anymore 🤷 

Can you update the rule-doc-generator and generate it again?

peterfox:
@driftingly I've done as asked.

PR #174, opened 2024-01-28
[Fix] Docs regenerated

peterfox:
Some docs appear to be wrong since #173


driftingly:
Thanks @peterfox probably my fault when I fixed merge conflicts on that last PR.  Thanks for catching and fixing.

PR #175, opened 2024-02-05
[CodeQuality] Adds rules for the with and value helpers

peterfox:
# Changes

* Adds a rule for removing redundant with() helper usage
* Adds a rule for removing redundant value() helper usage
* Adds tests for both new rules
* Updates rule list

# Why

Based on rules from Larastan. These are quite simple little use cases where we can remove the use of a helper when it's not needed.

```diff
-value(new Object())->something();
+(new Object())->something();
```

-with(new Object())->something();
+(new Object())->something();

value() => https://github.com/larastan/larastan/blob/2.x/src/Rules/UselessConstructs/NoUselessValueFunctionCallsRule.php
with() => https://github.com/larastan/larastan/blob/2.x/src/Rules/UselessConstructs/NoUselessWithFunctionCallsRule.php

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1,4 +1,4 @@
-# 51 Rules Overview
+# 53 Rules Overview
 
 ## AddArgumentDefaultValueRector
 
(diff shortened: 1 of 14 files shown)


driftingly:
Thanks @peterfox 🙌

PR #182, opened 2024-02-19
Dispatch to DispatchSync with non queueable job

peterfox:
# Changes

* Adds a rule to covert dispatch helper and other dispatcher calls to use dispatch sync methods instead
* Adds tests
* Updates docs
* Adds the rule to the Laravel 10

# Why

Performs the following:

```diff
-$result = dispatch(new SomeJob());
-$anotherResult = Bus::dispatch(new SomeJob());
-$anotherResult = $this->dispatch(new SomeJob());
+$result = dispatch_sync(new SomeJob());
+$anotherResult = Bus::dispatchSync(new SomeJob());
+$anotherResult = $this->dispatchSync(new SomeJob());
```

In Laravel 9 and earlier if you were able to send a non-queue job to dispatch and it would handle it immediately but now it will return a PendingBatch instead of the return value of the job. Even then it doesn't make sense to use dispatch instead of dispatchSync if it's a non queueable job so this should be helpful to a few people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,17 +17,18 @@ composer require driftingly/rector-laravel --dev
 To add a set to your config, use `RectorLaravel\Set\LaravelSetList` class and pick one of the constants:
 
 ```php
-use RectorLaravel\Set\LaravelSetList;
-use Rector\Config\RectorConfig;
+<?php
 
-return static function (RectorConfig $rectorConfig): void {
-    $rectorConfig->sets([
+declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
+use RectorLaravel\Set\LaravelSetList;
+
+return RectorConfig::configure()
+    ->withSets([
         LaravelSetList::LARAVEL_110
     ]);
-};
 ```
-
-<br>
 
 ## Learn Rector Faster
 
--- a/build/composer-php-72.json
+++ b/build/composer-php-72.json
@@ -5,7 +5,7 @@
     "description": "Rector upgrades rules for Laravel Framework",
     "require": {
         "php": "^7.2 || ^8.0",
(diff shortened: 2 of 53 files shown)


peterfox:
@GeniJaho I looked at this but it's very difficult to work out all possible contexts of a return value being used

The main problem with dispatch returning PendingBatch is that it uses a destructor so in some cases it will be fine. But the worst is when a variable is created so I just focused on that.

GeniJaho:
I was thinking, since we're only doing a rename, we could target all use cases by targeting the three nodes directly:
```php
public function getNodeTypes(): array
{
    return [FuncCall::class, MethoCall::class, StaticCall::class];
}
```

This would also simplify the rule, but I'm not sure if we should touch all cases anyway.
Please generate the docs once more, since your other PRs were merged.

peterfox:
@GeniJaho docs have been updated 👍 

In theory, there's no reason why it can't just be for any call I guess, as I can't see a reason why someone would dispatch an instance that didn't have ShouldQueue implemented.

driftingly:
Thanks @peterfox and @GeniJaho 
If you were using `dispatch` in an `if` or `foreach` I don't think you're expecting a `PendingBatch`. I'm leaning towards renaming all instances, not just assignments.

peterfox:
@GeniJaho @driftingly I've updated to refactor all calls rather than just assigned ones.

driftingly:
Thanks @peterfox 
Looks good to me 👍

peterfox:
@GeniJaho you okay with the rule in the current state?

GeniJaho:
@peterfox I was ok with the rule since 1 week ago 😁

PR #184, opened 2024-02-22
Declare type configs

peterfox:
Still a work in progress but would love some early feedback.

# Changes

* deprecates two older rules for the new type declaration sets.
* sets up a dir for tests for the type declaration sets.
* adds new type declaration configs, one for eloquent, one for service container.

# Why

One of the powerful things about Rector is being able to add type declarations. Laravel uses a lot of closures that users of the framework often won't type hint due to not knowing the correct types to apply. These configs make use of the rules I created in the base rector repo. This will allow for configs that define what types to apply to closures.

So far documented common eloquent type hints and service container ones.

# Notes

Currently pointed at rector dev-main until a new release with the closure type hinting exists.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "description": "Rector upgrades rules for Laravel Framework",
     "require": {
         "php": ">=8.2",
-        "rector/rector": "^1.0"
+        "rector/rector": "main-dev"
     },
     "require-dev": {
         "phpunit/phpunit": "^10.5",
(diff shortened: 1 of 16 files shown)


driftingly:
Thanks @peterfox 🙌 
I like where this is going.

peterfox:
Not giving up on this. But it's so out of date it would be hard to continue from this branch. Instead I need to look at the configurable rules available in Rector itself to make better decisions and reduce a lot of boiler plate

PR #185, opened 2024-02-25
Allow RouteActionCallableRector to work with \Route calls

### LINKED ISSUE #179: Using \Route fails silently with StaticCall\RouteActionCallableRector
`RouterRegisterNodeAnalyzer::isRegisterMethodStaticCall()`  expects `$node->class` to be `Illuminate\Support\Facades\Route`, but it can in fact be just `Route` (`\Route`).

This can be resolved by adding `use Illuminate\Support\Facades\Route;` in the routes file, but it's not obvious.

GeniJaho:
Fixes #179.


driftingly (approved):
Thanks @GeniJaho

PR #189, opened 2024-02-29
Adds ReplaceServiceContainerCallArgRector

peterfox:
# Changes

Adds a new ReplaceServiceContainerCallArgRector rule
Adds a ReplaceServiceContainerCallArg value object
Adds tests
Updates the docs

# Why

A neat little rule here which can be used in a set in the future to remove uses of string alias' to the service container.

```diff
-app('encrypter')->encrypt('...');
-\Illuminate\Support\Facades\Application::make('encrypter')->encrypt('...');
+app(Illuminate\Contracts\Encryption\Encrypter::class)->encrypt('...');
+\Illuminate\Support\Facades\Application::make(Illuminate\Contracts\Encryption\Encrypter::class)->encrypt('...');
```

Configuration of the rule is as simple as:

```php
new ReplaceServiceContainerCallArg(
    'encrypter',
    new ClassConstFetch(
        new FullyQualified('Illuminate\Contracts\Encryption\Encrypter'),
        'class'
    ),
),
```

You can also do it the other way around if you for some reason want to swap a Class const for a string.

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1,4 +1,4 @@
-# 56 Rules Overview
+# 57 Rules Overview
 
 ## AddArgumentDefaultValueRector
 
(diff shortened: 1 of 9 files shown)


driftingly:
This is great. Thanks @peterfox 
Feel free to merge when ready 🚀

PR #192, opened 2024-03-02
Adds a set to refactor service names to interfaces/classes

peterfox:
# Changes

* Adds a new set to apply the new rule from #189 to a set of possible string values found within the Laravel framework by default
* Adds a basic test covering a few of the classes

# Why

This will make it easier for people to clean up an app with the old-style strings used and replace them with appropriate references to interfaces or if not an interface, the class itself.

# Notes

The only problem I can see currently is there are no docs for these and this is a problem for #184 as well, people won't know these exist without extra documentation. I don't know if an extra markdown file that's just manually updated will do for now.


driftingly:
We could call out extra features in the readme rather than a separate md file. 
More likely to be used/updated if in the readme.

PR #194, opened 2024-03-04
Refactor If statements to use helpers with conditionals and provides a set

peterfox:
# Changes

* Adds rules for `abort_if` and `report_if` function calls as replacement for If statements containing `abort` or `report`.
* Adds tests for both rules
* Adds a set to handle the two new rules and the previous `throw_if` related rule. The set making it easier to just use all of them.
* Adds the `config` dir to rector

# Why

Makes sense to cover all the possible helpers and have a set that covers the three scenarios. The only one missing would be the Dispatchable trait static calls (dispatchIf) but it would require a new rule and feels a bit more niche compared to the helpers.


johnbacon:
I may be missing something, but this doesn't seem to be available via https://github.com/driftingly/rector-laravel/blob/eefaf3ef1186a4f07b50988de402b30b87143ed7/src/Set/LaravelSetList.php. Is there another way to use it? (Thank you!!)

GeniJaho:
@johnbacon You're right, we're missing a constant for it. We'll probably add one soon. 
Another way to use it is to include the three rules manually:
```php
$rectorConfig->rule(AbortIfRector::class);
$rectorConfig->rule(ReportIfRector::class);
$rectorConfig->rule(ThrowIfRector::class);
```

PR #196, opened 2024-03-09
Add nikic/php-parser and sort packages

peterfox:
Currently, the build fails because installing the packages allows for `nikic/php-parser` to install the latest version (^5.0) but currently the new API for v5 breaks PHPStan but resolving the PHPStan issues will cause the tests to fail. For now, locking the dev requirement keeps the build working as expected and doesn't appear to affect the running of tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,14 +8,15 @@
         "rector/rector": "^1.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^10.5",
-        "phpstan/phpstan": "^1.10",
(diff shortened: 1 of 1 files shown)


driftingly:
Thank you so much! I took a crack at this on friday, but couldn't figure out the issue.  I don't think I would have figured it out.

peterfox:
@driftingly it's not obvious right away. I think there will need to be fixes made in different repositories before we can make use of the v5 Parser

PR #206, opened 2024-04-20
Bugfix when queueing closures they were mistaken as non-queable jobs

### LINKED ISSUE #205: Rule DispatchNonShouldQueueToDispatchSyncRector will wrongly convert  `dispatch` to `dispatch_sync` on Closures
Laravel supports [queuing closures](https://laravel.com/docs/11.x/queues#queueing-closures) using `dispatch()`. Currently the rule `DispatchNonShouldQueueToDispatchSyncRector` would wrongly convert those `dispatch` calls to `dispatch_sync` .

Example:
```diff
- dispatch(function () {
+ dispatch_sync(function () {
    info('test');
});
```

#206 should fix that.

j3j5:
Fixes #205 and add a test for it.


j3j5:
Oh! I just realized that `Bus::dispatch()` [does](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Bus/Dispatcher.php#L76) [not](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Bus/Dispatcher.php#L205) in fact queue the closure, only [using the helper](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Foundation/helpers.php#L407) the closure gets queue. I wonder if I should report this upstream to the laravel repo, wdty

j3j5:
I've just changed it for now to match the current behaviour on the framework, that is, using `dispatch()` with a Closure will get the closure queued, using `Bus::dispatch()` will get it executed. I'm going to open an issue upstream and see if it's something they overlooked or is expected behaviour. If they fix it, I can resubmit another PR later.

GeniJaho:
@j3j5 what a find :grin:. It looks like undocumented behavior, it may not be a bug but as a user I would expect them to behave the same way. In any case, if this changes later and the Bus call is updated to queue closures, this Rule will need to be updated as well. In that regard I think it would be better if we skipped the Bus facade altogether for now, so as not to break existing/future code. If the behavior changes, we would probably add a rule in a version constraint, like from Laravel 11 and up only. What do you think?

j3j5:
So, as of now, my PR makes the rule to change all `Bus::dispatch()` calls to `Bus::dispatchSync()` (the calls which contain  something that's not implementing the `ShouldQueue` interface), which is exactly what's happening before my PR, right now it will only skip calls using the helper. The behaviour, as seen on the test is:

```diff
// Before
- Bus::dispatch(fn() => 'hello');
dispatch(fn() => 'hello');
---
// After using DispatchNonShouldQueueToDispatchSyncRector
+ Bus::dispatchSync(fn() => 'hello');
dispatch(fn() => 'hello');

```

j3j5:
Also, if you also think it's a bug, please, feel free to comment on the other issue, haha.

GeniJaho:
Sounds good, let's merge it :v:

PR #209, opened 2024-04-25
Update README.md with available sets

GeniJaho:
Added a new section in the Readme, since it's not obvious what the available Laravel sets are, what they do, and why.
Descriptions and links are taken from the comments in the rules/sets themselves, but we may need more descriptive paragraphs. @driftingly @peterfox what do you think?

image

I've also added two new constants to `LaravelSetList`, since we have the sets but they [can't be used](https://github.com/driftingly/rector-laravel/pull/194#issuecomment-2067703098) at all without their constants:
```php
LaravelSetList::LARAVEL_CONTAINER_STRING_TO_FULLY_QUALIFIED_NAME,
LaravelSetList::LARAVEL_IF_HELPERS
```

Sets were added in https://github.com/driftingly/rector-laravel/pull/192 and https://github.com/driftingly/rector-laravel/pull/194.


peterfox (approved):
Thanks @GeniJaho, this looks good to me.

I did float [a solution](https://github.com/rectorphp/rector/discussions/8535) for this in the main Rector repo but just haven't had the time/energy. It would be good to automate it all but for now I think this displays the options nicely.

GeniJaho:
The table in the Readme is quite bigger than I thought, will fix it soon :sweat_smile:.

Issue #215, opened 2024-05-06
dispatchNow is not converted to dispatchSync


jeff1326:
When i run rector laravel with this configuration, the dispatchNow is not replaced by dispatchSync.

In rector.php
```
return RectorConfig::configure()
    ->withPaths([
        __DIR__ . '/app',
        __DIR__ . '/config',
        __DIR__ . '/database',
        __DIR__ . '/modules',
        __DIR__ . '/public',
        __DIR__ . '/resources',
        __DIR__ . '/tests',
    ])
    ->withSets([
        LevelSetList::UP_TO_PHP_83,
        LaravelLevelSetList::UP_TO_LARAVEL_100,
    ]);
```
(I tried with PHPUnitSetList::PHPUNIT_100 too)

In my code
```
// Arrange
$product = Product::factory()->create();
$user = auth()->user();
        
// Act
ExportProduct::dispatchNow($product->id, $user->email);

// Assert
```
The line with `Product::factory()` has been changed by rector in this file, so it's parsed

PR #219, opened 2024-05-07
Add missing renames for dispatchNow method on Jobs

GeniJaho:
Fixes https://github.com/driftingly/rector-laravel/issues/215.


peterfox:
Have you tested this? It might not work as Dispatchables is a trait if I remember correctly and won't be picked up in the same way PHPStan does with Classes and Interfaces because it's not a type.

GeniJaho:
@peterfox Yes, both in a real project and with tests in the rector-laravel code. The `RenameMethodRector` takes trait uses into account when we provide traits instead of regular classes to it.

It would be great if we could add tests for the sets instead of just the rules. Not sure if this has been mentioned or considered before in the core Rector repo.

peterfox:
@GeniJaho okay good to know, I know when I've used PHPStan directly I found it didn't find traits so wanted to double check.

[I created a folder for testing Sets](https://github.com/driftingly/rector-laravel/tree/main/tests/Sets/ContainerStringToFullyQualifiedName) if you're asking if it's possible? It's just creating a config that points to the sets. Tests are otherwise the same as rule tests.

GeniJaho:
@peterfox that's exactly what I was looking for, thank you. I didn't know we already had them.

I added tests for the Laravel 10 Set and fixed a 'bad' configuration for renaming `$routeMiddleware` in the Http Kernel.

Issue #220, opened 2024-05-09
laravel90 set unnecessarily adding null argument to Model::touch()


hackel:
When I run the laravel90 set against my code, it is changing all instances of `$model->touch()` to `$model->touch(null)` for no apparent reason (null is the default value).  Is there a way to prevent this?  It doesn't seem possible to skip this particular change from the config.  Thanks!

PR #221, opened 2024-06-02
Fix redundant default arguments in Laravel 9 Set

GeniJaho:
Fixes https://github.com/driftingly/rector-laravel/issues/220.

This PR also handles similar cases in the Laravel 9 set, where we add redundant default variables, and covers with tests all the changes of the set.


peterfox (approved):
All looks good to me @GeniJaho 🎉

PR #222, opened 2024-06-22
Livewire set and first rule for QueryString to Url attributes

peterfox:
# Changes

- adds a Livewire 3.0 set to the repository.
- adds a new rule for Livewire to convert query strings to using Url attributes.

# Why

Livewire 3.0 is an official package for Laravel and a popular one. Livewire 3.0 introduced some changes to components from past versions. It makes sense for those rules to still live within the Laravel package.

# Example of Rule use

The rule will perform a simple conversion. It avoids applying if there's a mismatch between what's in the array and the properties existing.

```php
use Livewire\Component;

class MyComponent extends Component
{
    public string $something = '';

    public string $another = '';

    protected $queryString = [
        'something',
        'another',
    ];
}
```

becomes

```php
use Livewire\Component;

class MyComponent extends Component
{
    #[\Livewire\Attributes\Url]
    public string $something = '';

    #[\Livewire\Attributes\Url]
    public string $another = '';
}
```


src/Set/Packages/Livewire/LivewireLevelSetList.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace RectorLaravel\Set\Packages\Livewire;
+
+use Rector\Set\Contract\SetListInterface;
+
+final class CashierLevelSetList implements SetListInterface
```
MrPunyapal, line 9: I think you missed rename here 🤔

peterfox: Thank you @MrPunyapal , I've fixed that now

MrPunyapal:
what will happen in this case? 

```php 
public $foo;
public $search = '';
public $page = 1;

protected $queryString = [
    'foo',
    'search' => ['except' => ''],
    'page' => ['except' => 1],
];
```

peterfox:
@MrPunyapal currently it will ignore it. Do you know what the translation of that would be with the attribute? I forgot that was an option available

MrPunyapal:
That feature is no longer available.

So first do key to value.

```diff
- 'property' => ['except' => 'something'],
+ 'property' 
```

peterfox:
Okay cool. That should be simple enough to resolve 👍

peterfox:
@MrPunyapal I've now adapted for the scenario mentioned where the array's key might be the property.

MrPunyapal:
I guess we missed something? 🤔

The whole queryString property isn't deprecated 🤔


https://livewire.laravel.com/docs/upgrading#url-query-string

peterfox:
Doesn't really matter if it's not deprecated. The rule is still useful for those who want to convert from Livewire 2 to 3 as it covers most of the scenarios. I guess the only thing that should be done is to ignore any instances where `'keep' => true` is used.

MrPunyapal:
@peterfox I haven't found much in documentation 😃

- Then after looking into code i think we need to handle more than one case? 🤔
 Attribute: https://github.com/livewire/livewire/blob/main/src%2FFeatures%2FSupportQueryString%2FBaseUrl.php#L12-L18

Legacy support: https://github.com/livewire/livewire/blob/main/src%2FFeatures%2FSupportQueryString%2FSupportQueryString.php#L20-L26

peterfox:
Okay cool, so it's probably best to create the attributes with the parameters if set.

MrPunyapal:
> @MrPunyapal currently it will ignore it. Do you know what the translation of that would be with the attribute? I forgot that was an option available

Yeah revert to this for now 🤔

peterfox:
@MrPunyapal new version of the rule. Now covers transforming cases of 'as' and 'except' to be converted with the attributes into arguments.

MrPunyapal (approved):

MrPunyapal:
looks good

GeniJaho (approved):

PR #225, opened 2024-06-28
Use FullyQualified for class instead of Name

megawubs:
This ensures a leading slash `\` is added to the namespace of the class that is newed up, ensuring the namespace is loaded from the root and not relative to the current namespace.

Currently this rector breaks my code as it changes the code into something that does not work.

```diff
@@ @@
     public function handleTenant(TenantWithDatabase $tenant): void
     {
         $tenant->run(function (): void {
-            CreateTenantBrandingAvatarsJob::dispatch()->onQueue(Queue::CENTRAL);
+            dispatch(new App\Central\Jobs\CreateTenantBrandingAvatarsJob())->onQueue(Queue::CENTRAL);
         });
     }
 }
    ----------- end diff -----------

Applied rules:
 * DispatchToHelperFunctionsRector
```

Note the missing `\` before `App`. This results in a namespace that can not be resolved. And broken tests

image

But with the leading slash, it's correct.

image

### DIFF
--- a/src/Rector/StaticCall/DispatchToHelperFunctionsRector.php
+++ b/src/Rector/StaticCall/DispatchToHelperFunctionsRector.php
@@ -128,7 +128,7 @@ final class DispatchToHelperFunctionsRector extends AbstractRector
         return new FuncCall(
             new Name($method),
             [
-                new Arg(new New_(new Name($class), $staticCall->args)),
+                new Arg(new New_(new Name\FullyQualified($class), $staticCall->args)),
             ],
         );
     }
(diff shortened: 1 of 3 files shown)


megawubs:
I've tried running `composer fix-cs` but this changes a lot of code that I did not touch. I don't know if it's required before merge?

TomasVotruba:
Indeed, this is best practice to work with names 👍

GeniJaho:
@megawubs Thanks for the PR and sorry for the late reply. We've just fixed the code style issue and can merge your PR. Can  you please run `composer fix-cs` once more?

megawubs:
I've just updated the code and ran `composer fix-cs`

PR #237, opened 2024-07-31
Improve documentation about sets

spawnia:
https://github.com/driftingly/rector-laravel/pull/234#discussion_r1697277741

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,22 +25,34 @@ To add a set to your config, use `RectorLaravel\Set\LaravelLevelSetList` and pic
 Sets for higher versions include sets for lower versions.
 
 ```php
-return Rector\Config\RectorConfig::configure()
+<?php declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
+use RectorLaravel\Set\LaravelLevelSetList;
+
+return RectorConfig::configure()
     ->withSets([
-        RectorLaravel\Set\LaravelLevelSetList::LARAVEL_110,
+        LaravelLevelSetList::UP_TO_LARAVEL_110,
     ]);
 ```
 
+The sets in `RectorLaravel\Set\LaravelSetList` only contain changes related to a specific version upgrade.
+For example, the rules in `LaravelSetList::LARAVEL_110` apply when upgrading from Laravel 10 to Laravel 11.
+
 ## Additional Sets
 
-The other sets do not correlate directly to Laravel upgrades, but can be used to improve different aspects of your code.
-Use `RectorLaravel\Set\LaravelSetList` to add sets.
+To improve different aspects of your code, use the sets in `RectorLaravel\Set\LaravelSetList`.
 
 ```php
+<?php declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
+use RectorLaravel\Set\LaravelSetList;
+
 return Rector\Config\RectorConfig::configure()
     ->withSets([
-        RectorLaravel\Set\LaravelSetList::LARAVEL_CODE_QUALITY,
-        RectorLaravel\Set\LaravelSetList::LARAVEL_COLLECTION,
+        LaravelSetList::LARAVEL_CODE_QUALITY,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,24 +18,31 @@ Install the `RectorLaravel` package as dependency:
...
 
-To add a set to your config, use `RectorLaravel\Set\LaravelSetList` class and pick one of the constants:
+To add a set to your config, use `RectorLaravel\Set\LaravelLevelSetList` and pick the constant that matches your target version.
+Sets for higher versions include sets for lower versions.
 
 ```php
-<?php
+return Rector\Config\RectorConfig::configure()
+    ->withSets([
+        RectorLaravel\Set\LaravelLevelSetList::LARAVEL_110,
```
GeniJaho, line 29: The `LaravelLevelSetList` class only has constants like `UP_TO_LARAVEL_*`, so change the line to `RectorLaravel\Set\LaravelLevelSetList::UP_TO_LARAVEL_110` please.

Also, keep the rest of the code example the same, since it provides context and shows the `rector.php` file is very minimal. So bring back this part please:
```php
<?php

declare(strict_types=1);

use Rector\Config\RectorConfig;
use RectorLaravel\Set\LaravelSetList;

return RectorConfig::configure()
```

README.md
```diff
@@ -18,24 +18,31 @@ Install the `RectorLaravel` package as dependency:
...
+    ]);
+```
 
-declare(strict_types=1);
+## Additional Sets
 
-use Rector\Config\RectorConfig;
-use RectorLaravel\Set\LaravelSetList;
+The other sets do not correlate directly to Laravel upgrades, but can be used to improve different aspects of your code.
```
GeniJaho, line 35: This is not necessarily true, since sets like `LaravelSetList::LARAVEL_110` can directly help with upgrades, just without inheriting all the previous version changes. Maybe we can point that out?

README.md
```diff
@@ -18,24 +18,31 @@ Install the `RectorLaravel` package as dependency:
...
+Use `RectorLaravel\Set\LaravelSetList` to add sets.
 
-return RectorConfig::configure()
+```php
+return Rector\Config\RectorConfig::configure()
     ->withSets([
-        LaravelSetList::LARAVEL_110
+        RectorLaravel\Set\LaravelSetList::LARAVEL_CODE_QUALITY,
+        RectorLaravel\Set\LaravelSetList::LARAVEL_COLLECTION,
```
GeniJaho, line 42: Please use the same format as the existing version, with `declare` and imports:
```php
<?php

declare(strict_types=1);

use Rector\Config\RectorConfig;
use RectorLaravel\Set\LaravelSetList;

return RectorConfig::configure()
```
I think we can add here `LaravelSetList::LARAVEL_110`, on top of code quality and collection sets, to show that it exists apart from the `UP_TO_` sets.

README.md
```diff
@@ -18,24 +18,31 @@ Install the `RectorLaravel` package as dependency:
...
+        ...
     ]);
 ```
-## Available Sets
```
GeniJaho, line 46: We need some header for the next section, maybe we can call this `Available Custom Sets`? A better name is welcome :grin:

GeniJaho (changes requested):

README.md
spawnia: I think it fits under the previous header `## Additional Sets`. The code examples above explains the general usage, and subsequently all available sets are listed.

README.md
spawnia: I added an explanation in the previous section `Automate Laravel Upgrades`.

README.md
spawnia: I readded a shortened PHP opening tag and imports and fixed the constant.

README.md
spawnia: I changed this to be more accurate.

GeniJaho (approved):

PR #244, opened 2024-08-27
Adds the TypeHintTappableCallRector

peterfox:
# Changes

* Adds the new `TypeHintTappableCallRector` rule
* Adds tests for the rules
* Adds a stub for the Tappable trait

# Why

Should be a neat rule to help people automatically type hint the tap method.

```diff
-tap($collection, function ($collection) {}
+tap($collection, function (Collection $collection) {}
```

```diff
-(new Collection)->tap(function ($collection) {}
+(new Collection)->tap(function (Collection $collection) {}
```

### DIFF
--- a/src/Rector/FuncCall/TypeHintTappableCallRector.php
+++ b/src/Rector/FuncCall/TypeHintTappableCallRector.php
@@ -80,12 +80,16 @@ CODE_SAMPLE
         /** @var Closure $closure */
         $closure = $node->getArgs()[1]->value;
 
+        if ($closure->getParams() === []) {
+            return null;
+        }
+
         $this->refactorParameter($closure->getParams()[0], $node->getArgs()[0]->value);
 
         return $node;
     }
 
-    private function refactorParameter(Param $param, Node $node): bool
+    private function refactorParameter(Param $param, Node $node): void
     {
         $nodePhpStanType = $this->nodeTypeResolver->getType($node);
 
@@ -93,14 +97,12 @@ CODE_SAMPLE
         if ($param->type instanceof Node) {
             $currentParamType = $this->staticTypeMapper->mapPhpParserNodePHPStanType($param->type);
             if ($this->typeComparator->areTypesEqual($currentParamType, $nodePhpStanType)) {
-                return false;
+                return;
             }
         }
 
         $paramTypeNode = $this->staticTypeMapper->mapPHPStanTypeToPhpParserNode($nodePhpStanType, TypeKind::PARAM);
         $param->type = $paramTypeNode;
-
-        return true;
     }
 
     private function refactorMethodCall(MethodCall $methodCall): ?MethodCall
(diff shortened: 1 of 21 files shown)


src/Rector/FuncCall/TypeHintTappableCallRector.php
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        if (count($node->getArgs()) < 2 || ! $node->getArgs()[1]->value instanceof Closure) {
+            return null;
+        }
+
+        /** @var Closure $closure */
+        $closure = $node->getArgs()[1]->value;
+
+        $this->refactorParameter($closure->getParams()[0], $node->getArgs()[0]->value);
```
GeniJaho, line 83: I think we need an extra check here for the number of params. It's unlikely, but the rule would fail if it encounters a Closure without params: `tap('test', function () {});`.

src/Rector/FuncCall/TypeHintTappableCallRector.php
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        if (! $methodCall->getArgs()[0]->value instanceof Closure) {
+            return null;
+        }
+
+        /** @var Closure $closure */
+        $closure = $methodCall->getArgs()[0]->value;
+
+        $this->refactorParameter($closure->getParams()[0], $methodCall->var);
```
GeniJaho, line 119: We would need a check for the number of params here as well.

src/Rector/FuncCall/TypeHintTappableCallRector.php
```diff
@@ -0,0 +1,128 @@
...
+
+        /** @var Closure $closure */
+        $closure = $node->getArgs()[1]->value;
+
+        $this->refactorParameter($closure->getParams()[0], $node->getArgs()[0]->value);
+
+        return $node;
+    }
+
+    private function refactorParameter(Param $param, Node $node): bool
```
GeniJaho, line 88: This method's return value is never used. Maybe we could make it `void`?

tests/Rector/FuncCall/ThrowIfAndThrowUnlessExceptionsToUseClassStringRector/Fixture/fixture.php.inc
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace RectorLaravel\Tests\Rector\FuncCall\ThrowIfAndThrowUnlessExceptionsToUseClassStringRector\Fixture;
+namespace RectorLaravel\Tests\Rector\FuncCall\TypeHintTappableCallRector\Fixture;
```
GeniJaho, line 3: This namespace here is probably changed by mistake 😁 
The ones below as well.

peterfox:
Thank you for catching those @GeniJaho 😄

PR #249, opened 2024-09-12
Fixes duplicate entries and adds a test

peterfox:
# Changes

- Fixes the duplicate entries for the LaravelSetProvider
- Adds a unit test to check if it's providing the sets correctly.

# Why

Picked up as a mistake from #248


TomasVotruba:
LGTM :+1:

PR #252, opened 2024-09-17
Add ResponseHelperCallToJsonResponseRector

Stoffo:


src/Rector/MethodCall/ResponseHelperCallToJsonResponseRector.php
```diff
@@ -0,0 +1,62 @@
...
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Name\FullyQualified;
+use Rector\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @changelog https://laravel.com/docs/7.x/database-testing#creating-models
+ * @changelog https://laravel.com/docs/8.x/database-testing#instantiating-models
```
GeniJaho, line 17: Thanks for the PR :)

These two links are not related to the rule 😅

src/Rector/MethodCall/ResponseHelperCallToJsonResponseRector.php
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @see \RectorLaravel\Tests\Rector\FuncCall\FactoryFuncCallToStaticCallRector\FactoryFuncCallToStaticCallRectorTest
```
GeniJaho, line 19: This `@see` annotation needs to point to the correct test class.

src/Rector/MethodCall/ResponseHelperCallToJsonResponseRector.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * @param  MethodCall  $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->isName($node->name, 'json')) {
+            return null;
+        }
+
+        if (! $this->isName($node->var, 'response')) {
```
GeniJaho, line 56: This check is not sufficient, since it could work with cases like `$response->json()` when it's not an instance of `JsonResponse`. We should instead make this check for the `ObjectType`.

tests/Rector/FuncCall/RemoveRedundantWithCallsRector/Fixture/fixture.php.inc
```diff
@@ -2,7 +2,7 @@
 
 namespace RectorLaravel\Tests\Rector\FuncCall\RemoveRedundantWithCallsRector\Fixture;
 
-use RectorLaravel\Tests\Rector\FuncCall\RemoveRedundantWithCallsRector\Source\HelperObject;
+use RectorLaravel\Tests\Rector\MethodCall\ResponseHelperCallToJsonResponseRector\Source\HelperObject;
```
GeniJaho, line 5: This class is changed by mistake.

tests/Rector/FuncCall/RemoveRedundantWithCallsRector/Fixture/skips_calls_with_callables_as_a_second_arg.php.inc
```diff
@@ -2,7 +2,7 @@
 
 namespace RectorLaravel\Tests\Rector\FuncCall\RemoveRedundantWithCallsRector\Fixture;
 
-use RectorLaravel\Tests\Rector\FuncCall\RemoveRedundantWithCallsRector\Source\HelperObject;
+use RectorLaravel\Tests\Rector\MethodCall\ResponseHelperCallToJsonResponseRector\Source\HelperObject;
```
GeniJaho, line 5: This class is changed by mistake.

GeniJaho (changes requested):

src/Rector/MethodCall/ResponseHelperCallToJsonResponseRector.php
Stoffo: Sorry for that, typical copy + paste mistake 😅

src/Rector/MethodCall/ResponseHelperCallToJsonResponseRector.php
Stoffo: Good point! Didn't think of that. Do you know how to solve that issue? `$this->isObjectType($node, new ObjectType('Illuminate\Routing\ResponseFactory'))` is `false` in that case...

GeniJaho: The `response()` function returns an instance of `Illuminate\Contracts\Routing\ResponseFactory`, but it may be overkill to check the object type 😅 . It might suffice to just check that it's a FuncCall, with no arguments.

```php
if (! $node->var instanceof FuncCall) {
    return null;
}

if (count($node->var->args) > 0) {
    return null;
}
```

I'm curious, when/why would you use this rule?

GeniJaho: If you're going the ObjectType way, you'd have to use the `$node->var` in the check, like this `$this->isObjectType($node->var, new ObjectType('Illuminate\Routing\ResponseFactory'))`

tests/Rector/MethodCall/ResponseHelperCallToJsonResponseRector/config/configured_rule.php
```diff
@@ -0,0 +1,18 @@
...
+use Rector\Config\RectorConfig;
+use RectorLaravel\Rector\MethodCall\ResponseHelperCallToJsonResponseRector;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->import(__DIR__ . '/../../../../../config/config.php');
+
+    $rectorConfig->rule(ResponseHelperCallToJsonResponseRector::class);
+};
+
+function response(): ResponseFactory
```
GeniJaho, line 15: Remove this declaration here please. It might conflict with existing `response()` function in the apps, since it has no checks with `function_exists()`. We don't need it for tests to pass (when we're not using the ObjectType check).

peterfox:
It might be better to scale this up to be a rule to handle all response helper scenarios.

Or to make a new configurable rule for the main rector repo. `FuncCallToMethodCallRector` kind of does the same thing but it instead injects a property into a class via the constructor. I'd say a `MethodCallToNew` might make sense.

tests/Rector/MethodCall/ResponseHelperCallToJsonResponseRector/config/configured_rule.php
Stoffo: Yes I was aware that this was kind of hacky... Was just my first approach to test it. It's removed now

src/Rector/MethodCall/ResponseHelperCallToJsonResponseRector.php
Stoffo: @GeniJaho Thanks for your input, this fixed the issues!

 The reason I wrote this Rector is that I like this style better than using all the helper functions laravel provides to keep the code more "OOP Style" if you will. It's easier to follow what really happens at first glance.

GeniJaho:
> It might be better to scale this up to be a rule to handle all response helper scenarios.

I don't think it's possible to add them all under one configurable rule, I suspect other `response()` methods require more than just passing parameters to the `new` constructor. We could add them in other rules, and make a set out of it 👀 

> Or to make a new configurable rule for the main rector repo.

A new `MethodCallToNew` in the core repo would probably do the trick, but I'm not sure it can be done in a way that handles different use cases. We have a simple one here, just copy the method params into the class constructor, but how useful/common would that Rector be for other use cases? However, this makes me think we need some core helper rules for common transforms like these...

@peterfox so how do you want to handle this PR? 😅

peterfox:
@GeniJaho I think for now it's not a bad rule. Let's get it in and then if we find it doesn't scale we'll deprecate this for something else.

GeniJaho (approved):

PR #258, opened 2024-10-13
Make RemoveDumpDataDeadCodeRector rule configurable

jivanf:
This PR makes it possible to configure the `RemoveDumpDataDeadCodeRector` rule with an array of dump functions. `dd()` and `dump()` are removed by default if the rule isn't used with a configuration in order to keep the current behavior.

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -978,6 +978,8 @@ refactors calls with the pre Laravel 11 methods for blueprint geometry columns
 
 It will removes the dump data just like dd or dump functions from the code.`
 
+:wrench: **configure it!**
+
 - class: [`RectorLaravel\Rector\FuncCall\RemoveDumpDataDeadCodeRector`](../src/Rector/FuncCall/RemoveDumpDataDeadCodeRector.php)
 
 ```diff
--- a/src/Rector/FuncCall/RemoveDumpDataDeadCodeRector.php
+++ b/src/Rector/FuncCall/RemoveDumpDataDeadCodeRector.php
@@ -8,15 +8,22 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Stmt\Expression;
 use PhpParser\NodeTraverser;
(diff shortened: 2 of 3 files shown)


GeniJaho:
Thanks for the PR @jivanf. Will merge it after all the checks pass.

In addition to this, in a future PR, we should deprecate the rule entirely, since this can now be achieved with an existing core Rector rule:
```php
->withConfiguredRule(RemoveFuncCallRector::class, [
        'dd', 'dump', 'die',
])
```

jivanf:
Linting failed, but it should be good to go now.

jivanf:
Is there something wrong on my end? I was able to run the test for the rule, but it fails when I run all the tests:

```
Test Suite Started (RectorLaravel\Tests\Rector\Class_\AnonymousMigrationsRector\AnonymousMigrationsRectorTest, 3 tests)
Test Suite Started (RectorLaravel\Tests\Rector\Class_\AnonymousMigrationsRector\AnonymousMigrationsRectorTest::test, 3 tests)
Test Preparation Started (RectorLaravel\Tests\Rector\Class_\AnonymousMigrationsRector\AnonymousMigrationsRectorTest::test#0)
Assertion Succeeded (Constraint: is empty, Value: Array &0 [])
Before Test Method Called (RectorLaravel\Tests\Rector\Class_\AnonymousMigrationsRector\AnonymousMigrationsRectorTest::setUp)
Before Test Method Finished:
- RectorLaravel\Tests\Rector\Class_\AnonymousMigrationsRector\AnonymousMigrationsRectorTest::setUp
Test Prepared (RectorLaravel\Tests\Rector\Class_\AnonymousMigrationsRector\AnonymousMigrationsRectorTest::test#0)
PHP Fatal error:  Allowed memory size of 2147483648 bytes exhausted (tried to allocate 262144 bytes) in phar:///home/jivanf/development/web/forks/rector-laravel/vendor/phpstan/phpstan/phpstan.phar/src/Analyser/NodeScopeResolver.php on line 464
```

GeniJaho:
You should be OK on your side, it's probably some recent core Rector update that we should account for, and make a small release to be compatible with it.

PR #263, opened 2024-10-24
feat: adjust AddGenericReturnTypeToRelationsRector rule to be able to generate new generic style code

### LINKED ISSUE #259: AddGenericReturnTypeToRelationsRector not compliant with updated relation generic
In both Laravel (https://github.com/laravel/framework/pull/51851) as Larastan (https://github.com/larastan/larastan/pull/1990), the relation generics have been updated. Running the AddGenericReturnTypeToRelationsRector doesn't add the generic return types correctly anymore and even suggests to remove the correct ones. For example:

```diff
     /**
-     * @return \Illuminate\Database\Eloquent\Relations\HasMany<\App\Models\Post, $this>
+     * @return \Illuminate\Database\Eloquent\Relations\HasMany<\App\Models\Post>
      */
     public function posts(): HasMany
     {
          //
     }
```
I had a look at the implementation of the Rector rule but as of now I'm not confident enough how to fix the rule in the right way.

canvural:
This PR adjusts `AddGenericReturnTypeToRelationsRector` rule and makes it configurable to also work with new relation generics introduced in Laravel 11.15

Fixes #259


canvural:
Fixes #259

peterfox:
Making it a configurable rule is a bit of a pain as it will break people's Rector configs if they have the rule applied. A better approach would be to detect the Laravel version installed. I appreciate that there's not a class for this, but it wouldn't be too hard to implement.

GeniJaho:
If we have the `false` value as the default, it won't break existing configurations. However, users would need to know exactly when to update the Rector config to use the rule as a configurable one. The changes won't take effect otherwise.

I think we can merge the PR with only a change to give the configuration a default value of `false`, so existing configs won't break, and then figure out in another PR how to make the Rule check the version in Composer. These two PRs should probably be part of the same release. 👀

peterfox:
@GeniJaho I agree. Let's merge it.

The trick wouldn't be to use composer. I believe a const value exists in the Laravel framework's kernel we could use PHPStan to grab the version. Writing tests for this would be challenging. We'd have to consider how we do CI builds in a matrix to make it work.

canvural:
Yeah I thought about auto detecting the Laravel version, but couldn't find an easy way to do so. And not sure anymore if we even should do it. It's kinda users responsibility. 

> @GeniJaho I agree. Let's merge it.
> 
> The trick wouldn't be to use composer. I believe a const value exists in the Laravel framework's kernel we could use PHPStan to grab the version. Writing tests for this would be challenging. We'd have to consider how we do CI builds in a matrix to make it work.

Are you referring to [this](https://github.com/laravel/framework/blob/5527e72efe5d86cf83afe21944a47fe9a0bdc406/src/Illuminate/Foundation/Application.php#L48)? We'd need to detect the autoloader of the project and then ask PHPStan about this class. Maybe Rector itself has some helper class already for this case 🤔 

--- 

I updated the PR with making the configuration optional, and default to false.

peterfox:
Rector already loads the code via PHPStan. So you only need to use Reflection to grab the Version.

PR #266, opened 2024-10-27
Laravel 11 new generic types

GeniJaho:
Relates to https://github.com/driftingly/rector-laravel/pull/263.

- Makes the rule as non-configurable, as before, so users won't need to change their configurations.
- The check for the Laravel version is done within the rule, using Reflection, and checking the `VERSION` const of the Laravel `Application` class. Should we refactor this functionality into a class of its own, or wait for when we reach for it on future rules?
- I've used the Laravel Container's when-needs-give capabilities to provide an optional `Application` class for testing the two versions.
- I'm not sure if there is any performance penalty for doing that check on every model relationship. I don't think there is any significant penalty, but it's still something to remember.

Is there a better way to do this @peterfox?

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1,4 +1,4 @@
-# 70 Rules Overview
+# 71 Rules Overview
 
 ## AbortIfRector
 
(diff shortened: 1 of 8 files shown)


canvural:
I tested it in a project and it works great 👍🏽

PR #267, opened 2024-10-28
New Rule to use the `whereLike` clause in Laravel 11.x

GeniJaho:
Adds a new rule `WhereToWhereLike` (open for a better name) that will migrate all the `->where('name', 'like', '%Rector%')` clauses into the new `whereLike()` introduced in [Laravel 11.x](https://github.com/laravel/framework/pull/52147).
- The rule has an optional configuration for the Postgres driver, since it uses a 'case-sensitive' search by default.
- I would love to add it to the Laravel 11 and Code Quality sets, but we'd have to make a check in the rule for the Laravel version. In any case, it could cause issues for Postgres users if the rule is enabled by default, they'd have to configure it manually... and skip the one provided by the sets, somehow.


src/Rector/MethodCall/WhereToWhereLikeRector.php
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * @param  MethodCall  $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->isObjectType($node->var, new ObjectType('Illuminate\Database\Query\Builder'))) {
+            return null;
+        }
```
peterfox, line 87: You would be better off expanding this a little. Namely using the contract for the query builder instead of the object type.

src/Rector/MethodCall/WhereToWhereLikeRector.php
```diff
@@ -0,0 +1,180 @@
...
+                ),
+            ]);
+    }
+
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [MethodCall::class];
```
peterfox, line 77: This could be expanded to Static calls for a Model as well, as `Model::where('a', 'like', 'b')` wouldn't be covered by this rule as it uses a proxy system.

tests/Rector/MethodCall/WhereToWhereLikeRector/Fixture/Postgres/with_postgres_driver.php.inc
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace RectorLaravel\Tests\Rector\MethodCall\WhereToWhereLikeRector\Fixture;
+
+use Illuminate\Database\Query\Builder;
+
+class Fixture
```
peterfox, line 7: It's best not to name classes in tests the same. I've found it causes weird issues. It's best to name the class WithPostgresDriver to match the file name.

src/Rector/MethodCall/WhereToWhereLikeRector.php
```diff
@@ -0,0 +1,180 @@
...
+
+    public function configure(array $configuration): void
+    {
+        if ($configuration === []) {
+            $this->usingPostgresDriver = false;
+
+            return;
+        }
+
+        Assert::count($configuration, 1);
```
peterfox, line 121: this line seems a bit redundant as we know it's an array and the key is checked. It should affect anything if the config contains more keys

src/Rector/MethodCall/WhereToWhereLikeRector.php
```diff
@@ -0,0 +1,180 @@
...
+        Assert::keyExists($configuration, 'usingPostgresDriver');
+        Assert::boolean($configuration['usingPostgresDriver']);
+        $this->usingPostgresDriver = $configuration['usingPostgresDriver'];
```
peterfox, line 124: It might be useful to introduce a const for the array key that's public or just have a static method that creates the array e.g. `WhereToWhereLikeRector::usingPostgresDriver()` Then update the code sample for it.

peterfox (approved):
Great stuff @GeniJaho !

ibrunotome:
@GeniJaho Just a note for the ones who apply this rule, tests broke when applied this rule for the first where:

before rector (passing): 

```php
  $query
    ->where('name', 'ilike', DB::raw("UNACCENT('%{$filter}%')"))
    ->orWhereRelation('approaches.influencer', DB::raw('UNACCENT(searchable)'), 'ilike', DB::raw("UNACCENT('%{$filter}%')"));
```

after this rector rule ("ERROR: Object of class Illuminate\Database\Query\Expression could not be converted to string")

```php
  $query
    ->whereLike('name', DB::raw("UNACCENT('%{$filter}%')"))
    ->orWhereRelation('approaches.influencer', DB::raw('UNACCENT(searchable)'), 'ilike', DB::raw("UNACCENT('%{$filter}%')"));
```

PR #269, opened 2024-11-07
Updates export-ignores

raphaelstolt:
Added missing export-ignores.


talkinnl:
Removing LICENSE files as default behavior should not be OK. Sure, quite some engineers don't value it or find it boring, but licenses are a fact of life.

Also removing README... Quite a microoptimization, which kills any offline DX.

raphaelstolt:
The license information/type is still present in the composer.json file.

peterfox:
I'm unsure the background or purpose of this. Going to close it. Can be reopened with an explanation.

PR #277, opened 2024-12-11
Rector 2.0

GeniJaho:
Goes in line with the core Rector 2.0 release.
- Upgrades the rules to the latest Rector, PHPStan, and PHPParser versions
- Simplifies composer scripts
- Adds the `MigrateToSimplifiedAttributeRector` rule to the Laravel 9 set
- Removes the `pint.json` configuration, to keep it in line with the default config from Pint
- Adds a new `RectorLaravel\AbstractRector` class that replaces the core `AbstractRector` and implements the `DocumentedRuleInterface`. We still need the [removed interface](https://github.com/rectorphp/rector-src/pull/6477) because we generate docs from it.


rcerljenko:
here it is...  :)

https://github.com/rectorphp/rector/releases/tag/2.0.0

PR #280, opened 2024-12-17
AddGenericReturnTypeToRelationsRector fix for chained relation call

### LINKED ISSUE #274: AddGenericReturnTypeToRelationsRector doesn't work if there is a chained method
`AddGenericReturnTypeToRelationsRector` doesn't work if there is a chained method.

## Works
```php
/**
 * @return \Illuminate\Database\Eloquent\Relations\HasMany<\App\Models\Sale, $this>
 */
public function sales(): HasMany
{
    return $this->hasMany(Sale::class);
}
```

## Doesn't work

```php
public function latestSale(): HasOne
{
    return $this->hasOne(Sale::class)->latestOfMany();
}

public function client(): BelongsTo
{
    return $this->belongsTo(Client::class)->withTrashed();
}

public function invoices(): HasMany
{
    return $this->hasMany(Invoice::class)->orderBy('process_date', 'desc');
}
```

rudiedirkx:
Fixes #274

### DIFF
--- a/src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php
+++ b/src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php
@@ -265,6 +265,12 @@ CODE_SAMPLE
             return null;
         }
 
+        // Find deepest MethodCall, which is the first in code, to allow chaining:
+        // $this->hasMany(..)->orderBy(..)->with(..)
+        while ($methodCall->var instanceof MethodCall) {
+            $methodCall = $methodCall->var;
+        }
+
         // Called method should be one of the Laravel's relation methods
         if (! $this->doesMethodHasName($methodCall, self::RELATION_METHODS)) {
             return null;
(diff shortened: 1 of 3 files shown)


GeniJaho:
Thanks for the PR :smiley: Please add a test fixture for this case and we can merge it.

rudiedirkx:
What? Where? How? Etc.

GeniJaho:
😅  While this PR didn't break anything, we need a way to confirm it solves the issue. We need a test to prove that the changes work. 
There are test fixtures for the rule in `tests/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector/Fixture`.
Please add two more, one in `NewGenerics` and one in `OldGenerics`, call them `relation-with-chained-method-calls.php.inc`, or something similar.
If this is too much, we can merge the PR and I'll complete the rest, don't worry.

rudiedirkx:
What's the Old vs New distinction? The Laravel version, or PhpStan version, or other package version?

I think I got it working.

GeniJaho (approved):

GeniJaho:
It's the Laravel version. From 11.15.0 we need to use the new Generics, as the old ones do not work well anymore after that version.
Basically, it's about the second model in the generic annotation, instead of using the class name like `User`, we need to use `$this` from that version onwards.

PR #282, opened 2024-12-22
Adds ServerVariableToRequestFacadeRector rule

peterfox:
# Changes

- new ServerVariableToRequestFacadeRector rule
- adds tests
- updates docs

# Why

This converts a simple use of the `$_SERVER` PHP variable to use the Request facade.

```diff
-$_SERVER['VARIABLE'];
+\Illuminate\Support\Facade\Request::server('VARIABLE');
```

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1,4 +1,4 @@
-# 74 Rules Overview
+# 75 Rules Overview
 
 ## AbortIfRector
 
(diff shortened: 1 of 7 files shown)


GeniJaho:
@peterfox Are there any similar cases to the other PR of the `$_SESSION` global? I think we can include more cases in this rule, with a similar implementation.

peterfox:
Nothing I can think of. I need to look at [getallheaders](https://www.php.net/manual/en/function.getallheaders.php) but I'm not sure how 1:1 that is.

I think the only thing is possibly some key values such as:

```php
$_SERVER['HTTP_USER_AGENT'];
$_SERVER['REQUEST_METHOD'];
```

Would be better replaced by dedicated method calls. That might be better off with a dedicated rule though, so you'd apply this rule and then one to transform `\Illuminate\Support\Facade\Request::server('HTTP_USER_AGENT')` into `\Illuminate\Support\Facade\Request::userAgent()`.

GeniJaho (approved):

Issue #287, opened 2025-01-09
EloquentWhereTypeHintClosureParameterRector in booted method model


timk95:
The EloquentWhereTypeHintClosureParameterRector rector rule also applies to the creating method when applied in the booted method of a model. However this is not (necessarily) a query builder, it most likely is the model itself.

This results in a crash `Argument #1 ($model) must be of type Illuminate\Contracts\Database\Query\Builder, App\Models\DB\Model given`

Example:
```
    public static function booted(): void
    {
        static::creating(function ($model) {
            $model->id = Str::uuid();
            $model->created_at = now();
        });
    }
```
Becomes:
```
    public static function booted(): void
    {
        static::creating(function (\Illuminate\Contracts\Database\Query\Builder $model) {
            $model->id = Str::uuid();
            $model->created_at = now();
        });
    }
```

peterfox:
@timk95 are you sure it's definitely from this rule? I've tried to recreate this with tests in #290 but it doesn't happen. The [rule should only ever apply to `where` or `orWhere`](https://github.com/driftingly/rector-laravel/blob/main/src/Rector/MethodCall/EloquentWhereTypeHintClosureParameterRector.php#L108).

timk95:
@peterfox Yes, it is the same rule. See the dry run below.

```
1) app/Models/DB/Model.php:23

    ---------- begin diff ----------
@@ @@

     public static function booted(): void
     {
-        static::creating(function ($model) {
+        static::creating(function (\Illuminate\Contracts\Database\Query\Builder $model) {
             $model->id = Str::uuid();
             $model->created_at = now();
         });
    ----------- end diff -----------

Applied rules:
 * EloquentWhereTypeHintClosureParameterRector

```

What is strange is that in the main you mentioned the expectedObjectTypeAndMethodCall function is different from the one in 2.0.1. I looked in my local vendor map and the function is like this.

    ```
private function expectedObjectTypeAndMethodCall($node): bool
    {
        switch (true) {
            case $node instanceof MethodCall && $this->isObjectType(
                $node->var,
                new ObjectType('Illuminate\Contracts\Database\Query\Builder')
            ):
                return true;
            case $node instanceof StaticCall && $this->isObjectType(
                $node->class,
                new ObjectType('Illuminate\Database\Eloquent\Model')
            ):
                return true;
            default:
                return false;
        }
    }
```

[...]

PR #290, opened 2025-01-09
EloquentWhereTypeHintClosureParameterRector refactor

peterfox:
Related to #287 

# Changes

- modifies two rules where a downgrade was causing the rule to break.
- adds additional tests.
- refactor the tests to not define classes in the fixtures and use source objects instead.
- sets up the build mechanism to retest code after a downgrade.

# Why

It looks like the `DowngradeMatchToSwitchRector` rule has an issue there which ended up removing code required to check the method causing the downgraded `EloquentWhereTypeHintClosureParameterRector` to break.

@driftingly @GeniJaho, we should be careful to make sure contributors add skip fixtures to cover conditions that ensure the rules they build don't affect code they're not meant to.

### DIFF
--- a/.github/workflows/downgraded_release.yaml
+++ b/.github/workflows/downgraded_release.yaml
@@ -31,6 +31,9 @@ jobs:
             # copy PHP 7.2 composer
             -   run: cp build/composer-php-72.json composer.json
 
+            # run the tests against the downgraded rules
+            -   run: vendor/bin/phpunit tests
+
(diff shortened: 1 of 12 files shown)


peterfox:
I've fixed the downgrade rule that caused the problems https://github.com/rectorphp/rector-downgrade-php/pull/263

But it would still be good to get this merged in for the additional tests etc.

PR #291, opened 2025-01-10
Add UnaliasCollectionMethodsRector

spawnia:


GeniJaho:
Thanks for the PR @spawnia.
While the same can be achieved with the configurable rule `RenameMethodRector`, it would be difficult to include the `LARAVEL_COLLECTION` set and be able to skip a particular configured rule. So a specific rule like this makes sense in this case 👍.

Issue #292, opened 2025-01-13
ApplyDefaultInsteadOfNullCoalesceRector should ignore throws


shaffe-fr:
Hi,
`ApplyDefaultInsteadOfNullCoalesceRector ` should not be applied when the right side of the expression throws an exception.

Example:
```diff
- $baseUrl = config('services.google_maps.base_url') ?? throw new Exception('No Google maps base URL configured.');
+ $baseUrl = config('services.google_maps.base_url', throw new Exception('No Google maps base URL configured.'));
```
Regards

clivinn-shla81092:
Hello, we tried to solve the issue.

This is what we did:

Modified the ApplyDefaultInsteadOfNullCoalesceRector class to check if the right side of the coalesce operator is a throw expression. If it is, we skip the transformation to preserve the original behavior. Also added a new test case to verify this change.

You can review changes in this commit: https://github.com/clivinn-shla81092/driftingly-rector-laravel-292/commit/957be7cf6a493c8f0959dfdb2697dd23611d56c4.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

shaffe-fr:
The fix is working, but the following imports are not at the top of the file.
```
     use PhpParser\Node\Expr\Throw_;
     use PhpParser\Node\Stmt\Throw_ as ThrowStmt;
```

PR #294, opened 2025-01-14
Skip setting an exception as a default value in ApplyDefaultInsteadOfNullCoalesceRector

hans-thomas:
Hi, It will fix the setting an exception as a default value bug as reported in #292.


GeniJaho:
Thanks for the PR. This looks good, please run the linter and we can merge it.

PR #298, opened 2025-01-17
Add a command to generate new rules

GeniJaho:
# Why
It's very difficult to add new rules to the repository, every time you have to copy an existing one and change everything.

I think we should implement a way to generate the base scaffolding needed for writing a new rule.

Modifying the `CustomRuleCommand` in the core Rector repository would make it unnecessarily complicated (https://github.com/rectorphp/rector-src/pull/6676).

# Changes
- Ported an implementation of the original command without all the options and dependencies. The reason is I couldn't make `RectorConfig` run user-defined commands. I think they need to be configured in the core Rector implementation somehow.
- Added a new directory called `Commands`, registered in `autoload-dev` so it doesn't make it to the users' vendor.
- Added a new composer script to make it easy to run the command, called `composer make:rule`.
- Added two new dev dependencies to work with the filesystem, but they can be refactored out since the use-case is very limited.
- To Do: cleanup, remove dependencies, generate configurable rules, add to Readme, etc

@peterfox what do you think? Can we use something like this? Is there a better way to generate new rules?

### DIFF
--- a/commands/MakeRuleCommand.php
+++ b/commands/MakeRuleCommand.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+namespace RectorLaravel\Commands;
+
+use Nette\Utils\FileSystem;
+use Rector\Console\Command\CustomRuleCommand;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
+
+/**
+ * Modified version of
+ *
+ * @see CustomRuleCommand
+ */
+final class MakeRuleCommand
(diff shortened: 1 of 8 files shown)


peterfox:
@GeniJaho sorry I'd not seen this until now. I think this is great. Rector has a `custom-rule` command but it hard codes to paths and it's more designed for people adding rules to a project instead of things like this library. This seems to work well. There's maybe some UX things I'd change but I'd merge this for now so we can start getting the benefits from it.

peterfox (approved):

PR #305, opened 2025-02-20
Add new alias to `UnaliasCollectionMethods`

MrYamous:
Add new aliases to `UnaliasCollectionMethodsRector` added in #291

### DIFF
--- a/src/Rector/MethodCall/UnaliasCollectionMethodsRector.php
+++ b/src/Rector/MethodCall/UnaliasCollectionMethodsRector.php
@@ -70,7 +70,12 @@ CODE_SAMPLE
             $replacement = 'contains';
         } elseif ($this->isName($name, 'average')) {
             $replacement = 'avg';
-        } else {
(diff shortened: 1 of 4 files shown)


src/Rector/MethodCall/UnaliasCollectionMethodsRector.php
```diff
@@ -70,7 +70,12 @@ private function updateMethodCall(MethodCall $methodCall): ?MethodCall
             $replacement = 'contains';
         } elseif ($this->isName($name, 'average')) {
             $replacement = 'avg';
-        } else {
+        } elseif ($this->isName($name, 'unlessEmpty')) {
+            $replacement = 'whenNotEmpty';
+        } elseif ($this->isName($name, 'unlessNotEmpty')) {
+            $replacement = 'whenEmpty';
+        }
+         else {
```
peterfox, line 78: might want to correct this

```suggestion
        } else {
```

peterfox (approved):

PR #314, opened 2025-03-11
New rule OrWhereToWhereAny

GeniJaho:


GeniJaho:
While it looks like a promising rule, it gets so complex, and there are many variations to combine clauses for whereAny() and orWhereAny(), that this rule would cause more harm than benefits.

PR #318, opened 2025-03-20
`HelperFuncCallToFacadeClassRector`, allow with argument

parth391:
This pull request will change `resolve` helper to `App::get` facade.

And allow change to `App::make` facade, in case of more than 1 argument.


src/Rector/FuncCall/HelperFuncCallToFacadeClassRector.php
```diff
@@ -59,12 +59,12 @@ public function getNodeTypes(): array
...
+        if (! $this->isNames($node->name, ['app', 'resolve'])) {
             return null;
         }
 
         if (count($node->args) !== 1) {
-            return null;
+            return $this->nodeFactory->createStaticCall('Illuminate\Support\Facades\App', 'make', $node->args);
         }
 
         return $this->nodeFactory->createStaticCall('Illuminate\Support\Facades\App', 'get', $node->args);
```
peterfox, line 70: We have to be careful changing the logic because at current your code would change `app()` use to `App::make()`.

That said this rule is a bit broken and should be using `make` over `get` because they have slightly different behaviours in that `get` will throw and exception if the abstract does not exist and `make` just returns null. `app(<service>)` uses `make`.

I think the following is right:

```php
        if (count($node->args) > 0) {
            return $this->nodeFactory->createStaticCall('Illuminate\Support\Facades\App', 'make', $node->args);
        }

        return null;
```

There needs to be tests that cover skipping `app()` with no arguments if they don't exist already.

parth391: @peterfox thank you for pointing that out, i totally missed the no-argument scenario.

parth391:
@peterfox, i have updated the code and added the test case for no-argument.

PR #327, opened 2025-04-06
Downgrade to php 7.4 for release

samsonasik:
The build composer.json require rector 2,which require php 7.4


samsonasik:
@driftingly let's merge :+1:

PR #316, opened 2025-03-19
fix(eloquent relations): add pivot class for BelongsToMany relations

jrdnrc:
Add pivot class for BelongsToMany relations.

This is for the recent Laravel 12.3.0 update.

Apologies, I intended to open this PR against my own fork for testing, I can close this PR here if desired!

PR #328, opened 2025-04-08
Laravel App Analyzer

peterfox:
# Changes

- Adds an App analyzer that manages finding the version in Laravel
- Updates AddGenericReturnTypeToRelationsRector
- Updates AddGenericReturnTypeToRelationsRector tests
- Adds a InteractsWithLaravelVersion trait to use with tests
- Removes some files used to control app version within tests

# Why

Overall this is just a way to simplify how we make rules that might rely on the Laravel version.


GeniJaho (approved):
Great addition 💪

brunogaspar:
Hello guys, sorry to bother, but something on these changes is causing the following regression:

```diff
    ---------- begin diff ----------
@@ @@
     }

     /**
-     * @return BelongsTo<User, $this>
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo<\<REDACTED>\Users\Models\User, $this>
      */
     public function user(): BelongsTo
     {
    ----------- end diff -----------

Applied rules:
 * AddGenericReturnTypeToRelationsRector
```

I've downgraded to 2.0.3 (just to double check) and it did not happen there, then updated again to 2.0.4 and it yielded the above result sadly.

peterfox:
Just to clarify but is the only difference is that you're getting the fully qualified name instead of the short model?

Might be the PR before this that caused the break by not checking if the tag is already set.

brunogaspar:
Thanks for the reply @peterfox 

Yes exactly, it sets the FQCN all the time now. I use PHP CS Fixer to "remove" the FQCN but rector puts it back.

You're referring to https://github.com/driftingly/rector-laravel/pull/324 ? I can try to revert the changes locally to check

peterfox:
@brunogaspar no, more likely https://github.com/driftingly/rector-laravel/pull/316

brunogaspar:
ah i see..

i've reverted the file [to this version](https://github.com/jrdnrc/rector-laravel/blob/26e8d6e1a08191a0bb94e0e6b785daabc576fcd5/src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php) and the issue is gone, so you're correct, it's not your PR.

peterfox:
I've got a test that covers your scenario and confirmed it. I just need to do a bit of digging to fix it.

brunogaspar:
Thank you @peterfox !

PR #331, opened 2025-04-19
Stop overriding previous tags

peterfox:
# Changes

- AddGenericReturnTypeToRelationsRector updates
- additional test scenario for AddGenericReturnTypeToRelationsRector

# Why

Reported by @brunogaspar  that the rule was working okay but would overwrite tags with a fully qualified name even when the tag was already correct.

This PR seeks to change the situation so if the tag exists, then it will not apply any changes.

Looking at this rule. There's a few changes that might be required when it comes to the pivot changes that were added in #316. I have a feeling relationships like MorphToMany haven't been considered and it should probably assign the Pivot class if the relationship uses one.


peterfox:
@brunogaspar if you're able to check this branch out and try it, that would be very useful.

This should fix the issue you've been having.

brunogaspar:
@peterfox That seems to do it, thank you 🙏

GeniJaho (approved):

peterfox:
@GeniJaho I'm not getting that error but I'm getting others though.

PR #351, opened 2025-06-19
Allow the Factory rules to be configured with an allowlist

jradtilbrook:
This allows each of the factory rules to operate on an allow list of models.

The intention is to aid in adoption of the newer factories without having to apply it to a whole codebase at once, which can be troublesome on larger and/or legacy codebases.

Initially mentioned in https://github.com/driftingly/rector-laravel/discussions/350

PR #353, opened 2025-06-20
Add rule to add HasFactory to Models

jradtilbrook:
This adds a new rule that adds the `HasFactory` trait to Eloquent models if not present. Useful as part of the other factory related rules to upgrade pre-v8.0 code.

I think might be worth including this in the set and updating some docs - let me know 👍


peterfox:
Sorry @jradtilbrook things have been a bit busy lately. It looks overall good, I need to consider adding to the rule group and just the structure of the tests

peterfox:
@jradtilbrook would you be able to refactor your branch to structure your tests like you did in the other branch? #351

jradtilbrook:
👍 I'll try get onto that tonight

jradtilbrook:
@peterfox ready to go now

Tamas-hi:
Does it make sense to add HasFactory to all Models automatically / by default with the laravel-legacy-factories-to-classes set?

I still want to convert legacy factories to classes, but now all my models received a hasFactory trait, even those which do not a have factory at all.

jradtilbrook:
Hmm good question 👍 In my opinion, I think it does when used as part of the set. Since the set contains `FactoryDefinitionRector` to migrate the factory, it makes sense for it to also mark the model with `HasFactory` so its a complete migration to the newest standard for eloquent factories.

However, as it allows for configuration, if that does not work for you, you can use each part of the set separately and configure the allowlist of models on which it runs.

GeniJaho:
Upon second thought, I agree that keeping this rule out of the Legacy Factories set makes more sense.

I'm not sure that it's easy to configure a rule when it's part of a set. You either have to use it without configuration, or not use the set at all, and get the rules one by one. Either way, it's not a great experience. We might have 'broken' the set with the last release 😓 

Given we have other rules coming in for the Factories, we'll probably exclude this HasFactory rule from the legacy migration set, and include it in a new one, just for Factories.

PR #362, opened 2025-08-06
feat: add composer-triggered set for version sets

calebdw:
Hello!

Please see https://github.com/rectorphp/rector-src/pull/7110 and https://getrector.com/documentation/custom-set-provider#content-2-add-your-composertriggeredset-objects-to-the-provide-method for reference

These changes (with the PR mentioned above) will allow users to simply specify the following in their rector config and automatically have all the rectors up to the package version applied:

```php
    ->withComposerBased(laravel: true)
```

Thanks!


calebdw:
Thanks!

PR #363, opened 2025-08-10
feat: add ConvertEnumerableToArrayToAllRector

calebdw:
Hello!

This is a companion PR to https://github.com/larastan/larastan/pull/2311

The `toArray()` method on an Enumerable maps over all the items and recursively calls the `toArray()` method on each item if it is an Arrayable object. For collections that do not contain any Arrayable items, this is an unnecessary map operation and the method can be replaced with `all()` which simply returns the underlying array.

I also added the upstream `d` and `dd` functions to help when writing tests since those did not exist because they are rector dev dependencies.

Thanks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
         "phpunit/phpunit": "^10.5",
         "symplify/rule-doc-generator": "^12.2",
         "tightenco/duster": "^3.1",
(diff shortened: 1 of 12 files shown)


calebdw:
Thank you sir!

PR #364, opened 2025-08-10
feat: add ReplaceFakerPropertyFetchWithMethodCallRector

calebdw:
Hello!

Property fetch on Faker\Generator has been deprecated since version [1.14.0](https://github.com/FakerPHP/Faker/releases/tag/v1.14.0). It should be replaced with the method call.

Thanks!


config/sets/packages/faker/faker-114.php
GeniJaho, line 1: Although the property fetch is deprecated since 1.14, we don't need to have it in a separate set. This rule can be used regardless of the Faker version, since method calls work on older versions as well.
Can you remove this set, and the FakerSetList & Composer set below please? Also, I think we should update the docs with `composer docs` since we're adding a new rule.

calebdw: Sure, what set would you like it to go in?

GeniJaho: I'm not sure about that. We have lots of rules about Factories, some of them are part of upgrade sets like Laravel 9.0, some are part of the `LARAVEL_LEGACY_FACTORIES_TO_CLASSES` set, and some are not part of any set.
This could also fit in the Legacy Factories set somehow, but maybe we need another set entirely that is just for Factories.

GeniJaho: Yeah, 1.0 seems better, let's do that. They don't release breaking changes that often.
We can add existing Faker rules to the set later.

calebdw: @GeniJaho, I've updated the version and fixed and executed the `composer docs` command (some of the configured rules did not use the config code example)

PR #367, opened 2025-08-12
Fix `UseForwardCallsTraitRector`

guanguans:
> Correct `ForwardCalls` to `ForwardsCalls`.

## Related links

* [https://github.com/laravel/framework/blob/12.x/src/Illuminate/Support/Traits/ForwardsCalls.php#L8C7-L8C20](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Support/Traits/ForwardsCalls.php#L8C7-L8C20)
* [https://github.com/driftingly/rector-laravel/blob/main/src/Rector/Class_/UseForwardCallsTraitRector.php#L27C75-L27C87](https://github.com/driftingly/rector-laravel/blob/main/src/Rector/Class_/UseForwardCallsTraitRector.php#L27C75-L27C87)

## CI

* [Tests](https://github.com/guanguans/rector-laravel/actions/runs/16898445203)
* [Code Analysis](https://github.com/guanguans/rector-laravel/actions/runs/16898445208)

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1592,16 +1592,16 @@ Use `$this->components` property within commands
 
 <br>
 
-## UseForwardCallsTraitRector
+## UseForwardsCallsTraitRector
 
 Replaces the use of `call_user_func` and `call_user_func_array` method with the CallForwarding trait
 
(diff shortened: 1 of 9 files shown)


GeniJaho:
Thanks for the PR @guanguans. This is a breaking change unfortunately, as it will break the configurations for users who tried the rule this week, but the rule was wrong itself. We should've been more careful before merging 🙏

PR #368, opened 2025-08-12
feat: improve AppEnvironmentComparisonToParameterRector

calebdw:
Hello!

This adds several improvements:
- support for the `in_array` function
- support for `!=` and `!==` operators
- support for the cleaner `isLocal` and `isProduction` methods

I also saw that this rule was not in any set, so I created a code style set (similar to the upstream rector one) for this and others (I figured this was more stylistic than code quality)

Thanks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "symplify/rule-doc-generator-contracts": "^11.2"
     },
     "require-dev": {
-        "nikic/php-parser": "^5.3",
+        "nikic/php-parser": "^5.6",
         "phpstan/extension-installer": "^1.3",
         "phpstan/phpstan": "^2.0",
         "phpstan/phpstan-deprecation-rules": "^2.0",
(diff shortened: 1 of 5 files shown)


GeniJaho:
@calebdw Thanks for the PR 🙌 Please add this rule to the existing Code Quality set. We don't have many rules there, and for the existing ones there is no hard distinction between quality and style. We've also found it's already difficult for people to discover new rules, let alone new sets 😅
Why are we upgrading `"nikic/php-parser": "^5.6",`?
I have to check the rest of the code still.

calebdw:
> Why are we upgrading "nikic/php-parser": "^5.6",?

So that way I can take advantage of the new [`CallLike::getArg()` method](https://github.com/nikic/PHP-Parser/pull/1089) which allows me to remove this code

image

calebdw:
Thanks!

PR #369, opened 2025-08-12
feat: add rector to add JsonResponse return value

giacomomasseron:
This PR add `AddJsonResponseRector` rector.

### Example

Before

```
public function __invoke()
{
    return new JsonResponse(['key' => 'value']);
}
```

After

```
public function __invoke(): \Illuminate\Http\JsonResponse
{
    return new JsonResponse(['key' => 'value']);
}
```

### DIFF
--- a/src/Rector/Class_/AddJsonResponseRector.php
+++ b/src/Rector/Class_/AddJsonResponseRector.php
@@ -1,0 +1,116 @@
+<?php
+
+declare(strict_types=1);
+
+namespace RectorLaravel\Rector\Class_;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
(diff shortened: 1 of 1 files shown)


calebdw:
We shouldn't need a rector specifically for this---there's plenty of other rectors that can add the return type based on what's returned

giacomomasseron:
I did not find any. You know the name of the rector that does that?

calebdw:
- https://getrector.com/rule-detail/return-type-from-strict-typed-call-rector
- https://getrector.com/rule-detail/return-type-from-return-new-rector
- https://getrector.com/rule-detail/return-type-from-strict-typed-property-rector
- https://getrector.com/rule-detail/return-union-type-rector
- ... and plenty more

You need to use the upstream `typeDeclarations` set

PR #370, opened 2025-08-13
feat: add MakeModelAttributesAndScopesProtectedRector

calebdw:
Hello!

This automatically converts all Model attributes and scopes to protected. It is actually an error for them to be `private` (as they are called from the parent Model) and they should not be `public` as a user should never call them directly

Thanks!


calebdw:
Thanks!

PR #371, opened 2025-08-13
feat: add additional methods to laravel-arrayaccess-to-method-call

calebdw:
Hello!

This adds support for the exists, set, and unset operations that can happen on an ArrayAccess object.

This also adds the Cache contract to the list and cleans up the implementations---they are not needed since the contracts are listed and they all implement the contracts.

Thanks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "description": "Rector upgrades rules for Laravel Framework",
     "require": {
         "php": ">=8.3",
-        "rector/rector": "^2.0.0",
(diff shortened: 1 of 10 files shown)


GeniJaho:
Thanks for the PR @calebdw.
We have tests for the sets as well. Please update `tests/Sets/ArrayAccessToMethodCall/Fixture/fixture.php.inc` or add other fixtures in the same directory, for the new methods like `get`, `set` etc.

calebdw:
Updated :+1:

PR #372, opened 2025-08-13
docs: update docs for version auto-detection

calebdw:
Hello!

This updates the readme with the new configure instructions now that the package support composer based sets

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,22 @@ composer require --dev driftingly/rector-laravel
 
 ## Automate Laravel Upgrades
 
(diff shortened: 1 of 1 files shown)


GeniJaho:
Thanks for the PR @calebdw. We've had issues in the past with the way we recommend these sets to be used, or even discovered.
For that reason, we need to keep the existing section the same, the one that shows how you can use the Laravel Level sets 'manually'.
You can add your suggestions under the same header of 'Automate Laravel Upgrades', to show users that if you want to use these level sets, you can use these helpers etc.

calebdw:
Done :+1:

PR #374, opened 2025-08-18
Add a @use annotation to HasFactory trait usages

imliam:
This PR adds a new Rector rule that automatically adds `@use` docblock annotations to Laravel models using the HasFactory trait, providing better IDE support and type hints by detecting the relevant factory and passing generic data through to it.

```diff
  class User extends Model
  {
+     /**
+      * @use HasFactory<\Database\Factories\UserFactory>
+      */
      use HasFactory;
  }
```


tests/Rector/Class_/AddUseAnnotationToHasFactoryTraitRector/Fixture/deep_namespace.php.inc
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace App\Models\Tenant;
```
GeniJaho, line 3: Please use the correct namespace here, which would be `namespace RectorLaravel\Tests\Rector\Class_\AddUseAnnotationToHasFactoryTraitRector\Fixture;`.
To make the test work you would have to make the `use` annotation a bit different:
```
@use \Illuminate\Database\Eloquent\Factories\HasFactory<\RectorLaravel\Tests\Rector\Class_\AddUseAnnotationToHasFactoryTraitRector\Source\Tenant\UserFactory>
```
I know it does not look like the the real thing with `App\Models`, but wrong namespaces in tests have caused us issues in the past.

src/Rector/Class_/AddUseAnnotationToHasFactoryTraitRector.php
```diff
@@ -0,0 +1,198 @@
...
+use Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfoFactory;
+use Rector\Comments\NodeDocBlock\DocBlockUpdater;
+use RectorLaravel\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see \RectorLaravel\Tests\Rector\Class_\AddUseAnnotationToHasFactoryTraitRector\AddUseAnnotationToHasFactoryTraitRectorTest
+ */
+final class AddUseAnnotationToHasFactoryTraitRector extends AbstractRector
```
GeniJaho, line 23: This needs to be a configurable rule, so that we are able to pass it the factories namespace, which would default to `Database\Factories`.
This way we could test it as well.
You can check how it is done on other configured rules, for example `WhereToWhereLikeRector`, where the configuration is a boolean. In this case it would be a string like `FACTORIES_NAMESPACE` or similar.

GeniJaho (changes requested):
Thanks for the PR @imliam 🙌

calebdw:
Note that if the `$factory` property is defined then can just take the factory directly from that instead of trying to auto-discover it

src/Rector/Class_/AddUseAnnotationToHasFactoryTraitRector.php
```diff
@@ -0,0 +1,198 @@
...
+
+        $factoryClassNames = $this->getPotentialFactoryClassNames($currentNamespace, $factoryName);
+
+        foreach ($factoryClassNames as $factoryClassName) {
+            if ($this->reflectionProvider->hasClass($factoryClassName)) {
+                return $factoryClassName;
+            }
+        }
+
+        return null;
```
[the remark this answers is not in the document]
imliam: I kind of see what you mean but feel like this code is still needed to handle some edge cases after quickly playing with it. Happy to give it another shot if it's a deal breaker?

imliam:
I've pushed up changes that handle all (but one) of the comments above. It's now configurable, test source is in an appropriate namespace, and it checks for the factory property before trying to resolve the class itself.

PR #382, opened 2025-08-30
feat: add ArrayToArrGetRector

MrPunyapal:
This pull request introduces a new Rector rule that automatically converts direct array access to use Laravel's `Arr::get()` method, improving code consistency and maintainability when working with nested arrays. The implementation includes comprehensive support for both simple array access and null coalesce patterns, with smart handling of edge cases.

**New Rector Rule: Array Access to `Arr::get()` Conversion**

* Added `ArrayToArrGetRector` rule to convert array access patterns:
  - Simple access: `$array['key']` → `\Illuminate\Support\Arr::get($array, 'key')`
  - Nested access: `$array['nested']['key']` → `\Illuminate\Support\Arr::get($array, 'nested.key')`
  - With defaults: `$array['key'] ?? 'default'` → `\Illuminate\Support\Arr::get($array, 'key', 'default')`
  - Preserves throw expressions: `$array['key'] ?? throw new Exception()` (unchanged)

* The rule handles both scalar string and integer keys, building proper dot notation for nested access
* Skips dynamic keys, complex expressions, and null coalesce with throw statements to maintain safety and original behavior

### Example
```diff
-$array['key'];
-$array['nested']['key'];
-$array['key'] ?? 'default';
-$array['nested']['key'] ?? 'default';
+\Illuminate\Support\Arr::get($array, 'key');
+\Illuminate\Support\Arr::get($array, 'nested.key');
+\Illuminate\Support\Arr::get($array, 'key', 'default');
+\Illuminate\Support\Arr::get($array, 'nested.key', 'default');
$array['key'] ?? throw new Exception('Required');

### CODE REVIEW

**Comment by MrPunyapal** [comment]
What do you say? @peterfox

**Comment by Hackbard** [comment]
Great! Do it

**Comment by calebdw** [comment]
Not sure I really care for this rule for non-nested array access? Why nearly double the number of characters needed?

```diff
-$array['key'];
+data_get($array, 'key');
```

Now, converting a null coalesce could be useful:

```diff
-$array['key'] ?? null;
+data_get($array, 'key');
-$array['key'] ?? 'foo';
+data_get($array, 'key', 'foo');
```

MrPunyapal:
I'm open to making it for `Arr::get()`.

peterfox:
@MrPunyapal sorry for my slowness. Been busy of late. The rule looks really good. I agree, it might be better to use `Arr::get()` instead of `data_get` as `data_get` is designed to work with objects as well. I can see this being useful as it is and we can look at expanding it to handle coalescing for defaults as well.

There's some useful code here as well I'd like to put into their own classes but that can be done later once this is merged.

With that in mind, would you be okay to make the change to `Arr::get`?

MrPunyapal:
@peterfox updated things based on your suggestions.
let me know if you need any changes

wlkns:
I don't understand. Why introduce slowness and pull away from core PHP just for the sake of using Laravel methods?

Did anyone benchmark this on a large scale?

peterfox:
> I don't understand. Why introduce slowness and pull away from core PHP just for the sake of using Laravel methods?
> 
> Did anyone benchmark this on a large scale?

1. People don't have to use it
2. Some people might want to use it for reasons you might not get right away
3. I'd use this for some readability personally

PR #386, opened 2025-09-16
fix: ContainerBindConcreteWithClosureOnlyRector Rule when the abstract is a variable

jf-m:
Hi,

This is a proposed solution to solve the following issue when using `LARAVEL_120` set (`ContainerBindConcreteWithClosureOnlyRector` Rule):

The following piece of code
```php
foreach ([SomeClass::class, SomeOtherClass::class] as $item) {
    $app->singleton($item, function () use ($item) {
        return new $item();
    });
}
```

Currently gets updated wrongly into
```php
foreach ([SomeClass::class, SomeOtherClass::class] as $item) {
    $app->singleton($item, function () use ($item): SomeClass::class|SomeOtherClass::class {
        return new $item();
    });
}
```

[EDIT: The above is a made-up example to explain simply the issue, see my comment bellow]

I believe we should prevent any modification when the first argument is a variable. The above code should remain unchanged.

This PR is a proposed approach to do that and will fix this bug.

---

Thank you very much for creating this wonderful package ;) 

Hope this helps,
Jeff

### DIFF
--- a/src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
+++ b/src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
@@ -3,6 +3,7 @@
 namespace RectorLaravel\Rector\MethodCall;
 
 use PhpParser\Node;
+use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Const_;
 use PhpParser\Node\Expr\Closure;
 use PhpParser\Node\Expr\MethodCall;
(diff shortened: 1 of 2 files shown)


calebdw:
It's really better if every singleton has it's own line:

```php
$app->singleton(fn (): SomeClass => new SomeClass());
$app->singleton(fn (): SomeOtherClass => new SomeOtherClass());
```

Additionally, if all you're doing is just newing up the class then you can directly add the class strings to the `$singleton` property

jf-m:
Hi @calebdw

Thanks for your feedback. 
Actually, I was just showing a made-up piece of code as an example to show the bug in a "simple" way.


In my application, I am actually not using singleton but `->bind()` and I'm loading a list of class through a config file, then the content of the `bind()` is very specific to my application and isn't just a `new $class()`.

Again, it was just to explain in a simple way my issue 🤓

jf-m:
And to add more details about my use-case, since the mentionned config file is not directly accessible to rector, in my case it transforms 

```
$app->bind($class, function () {[..]}
```

into 

```
$app->bind(function (): object {[..]}
```

Which is even more of an issue (this PR solves this)

PR #387, opened 2025-09-16
fix: skip rule when its overriding the method

canvural:
This PR skips the `MakeModelAttributesAndScopesProtectedRectorTest` if the method is overriding a public parent method. Narrowing down the type in a child class results in a fatal error.


calebdw:
It is legal to change method visibility in a child class?

canvural:
You can widen but not narrow it down: https://3v4l.org/bDk2B

calebdw:
Ah that makes sense

PR #398, opened 2025-10-04
Convert model ID comparisons to use the is() method

imliam:
```diff
- $team->user_id === $user->id;
+ $team->user()->is($user);
```

### DIFF
--- a/src/Rector/Expr/ModelComparisonToIsMethodRector.php
+++ b/src/Rector/Expr/ModelComparisonToIsMethodRector.php
@@ -13,6 +13,7 @@ use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\PropertyFetch;
 use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Identifier;
+use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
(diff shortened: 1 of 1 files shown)


src/Rector/Expr/ModelComparisonToIsMethodRector.php
```diff
@@ -0,0 +1,150 @@
...
+
+    private function isForeignKeyToIdPattern(string $leftProperty, string $rightProperty): bool
+    {
+        return str_ends_with($leftProperty, '_id') && $rightProperty === 'id';
+    }
+
+    private function extractRelationshipName(string $foreignKeyProperty): string
+    {
+        return substr($foreignKeyProperty, 0, -3);
+    }
```
GeniJaho, line 124: I'm afraid this part here won't work well with a lot of cases.

It assumes that the key is always 'id' or follows the standard '_id'. I'm not sure if we can find the key name from the model's `getKeyName()` method or the `$primaryKey` property.

Also, it fails when relationship names are complex (`nationalTeam()` with `national_team_id` key), or don't exist (`external_team_id`, but no `externalTeam()` relationship).

The only way this could work well would be to check the relationships.

GeniJaho (changes requested):

PR #399, opened 2025-10-04
Refactor `config()` calls to use type-specific methods when the expected type is known

imliam:
```diff
- $name = (string) config('app.name');
- $lifetime = (int) config('session.lifetime');
- $debug = (bool) config('app.debug');
- $version = (float) config('app.version');
- $connections = (array) config('database.connections');

+ $name = config()->string('app.name');
+ $lifetime = config()->integer('session.lifetime');
+ $debug = config()->boolean('app.debug');
+ $version = config()->float('app.version');
+ $connections = config()->array('database.connections');
```


GeniJaho:
Lovely PR :raised_hands: As a follow up, we could also include calls like `config()->get()` to have the same behavior.
And in a next release we should put this rule under the Code Quality set, after users try it for a few weeks and we iron out any bugs.

calebdw:
Keep in mind that this changes behaviour, and throws exceptions if the value is  not the correct datatype rather than simply casting

PR #400, opened 2025-10-04
Refactor `abort_if()` calls to `Gate::denyIf()` when a user is concerned

imliam:
Per [this thread](https://x.com/michaeldyrynda/status/1958398521700470971), the reasons for preferring Gates here are:

- `abort_if()` throws a plain HttpException, while `Gate::denyIf` throws a more precise AuthorizationException which can be used by the exception handler
- Gates get their hooks picked up, so admins/etc. won't be tripped up by these guard clauses

```diff
- abort_if($user->id === $post->user_id);
- abort_if($post->user()->is($user));
+ Gate::denyIf($user->id === $post->user_id);
+ Gate::denyIf($post->user()->is($user));
```

This is a quite opinionated one so I wouldn't include it as part of any default sets, but still think it's an interesting and useful refactor.

### DIFF
--- a/src/Rector/FuncCall/AbortIfToGateDenyIfRector.php
+++ b/src/Rector/FuncCall/AbortIfToGateDenyIfRector.php
@@ -5,8 +5,8 @@ declare(strict_types=1);
 namespace RectorLaravel\Rector\FuncCall;
 
 use PhpParser\Node;
+use PhpParser\Node\Arg;
 use PhpParser\Node\Expr\FuncCall;
-use PhpParser\Node\Expr\MethodCall;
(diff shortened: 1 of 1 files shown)


GeniJaho:
I feel like this rule could remove some checks that are supposed to prevent all types of users, and there's no way to handle all cases correctly.

For example, `abort_if($user->posts()->count() < 100);`. This is not a user access permission check, but the rule would change it.

peterfox:
@imliam I like the rule in theory but as @GeniJaho pointed out, it's a bit problematic and I can't think of any way to reduce that. I think this could become one of those rules that just adds to maintenance work later on without providing a ton of value to end users. We appreciate the effort though.

PR #401, opened 2025-10-04
Use whereKey query method where possible

imliam:
```diff
- User::where('id', '=', $user->id)->get();
+ User::whereKey($user)->get();
```

### DIFF
--- a/src/Rector/MethodCall/EloquentWhereIdToWhereKeyRector.php
+++ b/src/Rector/MethodCall/EloquentWhereIdToWhereKeyRector.php
@@ -1,0 +1,165 @@
+<?php
+
+declare(strict_types=1);
+
+namespace RectorLaravel\Rector\MethodCall;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\PropertyFetch;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Scalar\String_;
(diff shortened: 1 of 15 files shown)


src/Rector/MethodCall/EloquentWhereIdToWhereKeyRector.php
```diff
@@ -0,0 +1,165 @@
...
+        $args = $node->getArgs();
+        $argCount = count($args);
+
+        if ($argCount === 2) {
+            return $this->refactorTwoArgumentWhere($node, $args, $isWhereNot);
+        }
+
+        if ($argCount === 3 && $isWhere) {
+            return $this->refactorThreeArgumentWhere($node, $args);
+        }
```
calebdw: can these be condensed into a single method? they look quite similar

src/Rector/MethodCall/EloquentWhereIdToWhereKeyRector.php
```diff
@@ -0,0 +1,165 @@
...
+        if ($columnName !== 'id') {
+            return null;
+        }
+
+        if (! $args[1] instanceof Arg || ! $args[1]->value instanceof PropertyFetch) {
+            return null;
+        }
+
+        $propertyFetch = $args[1]->value;
+        if (! $this->isName($propertyFetch->name, 'id')) {
```
calebdw: instead of hardcoding `id` here, it would be better to get a new model instance from the helpers, and then use the result of `->getKeyName()`

calebdw:
I'm not sure if there's another rule that does this, but I'd also like a rule (or just update this rule) that turns `whereIn('id', ...)` into `whereKey()` as well

PR #402, opened 2025-10-04
Refactor Carbon static calls to Date Facade

imliam:
Laravel's Date facade has some benefits; notably that it adheres to whatever handler is configured (so if you call `Date::use(CarbonImmutable::class)` in a service provider, all your dates will be immutable).

```diff
- Carbon::now();
+ Date::now();
```


Tamas-hi:
RectorPHP has a prepared set called "carbon" that converts datetime to carbon. 
I think we have a conflict here, and makes it harder for the end user which one to use.

PR #403, opened 2025-10-04
Refactor attempts to get request data to use type-specific methods when the type is known

imliam:
```diff
-$name = $request->input('name');
-$age = (int) $request->get('age');
-$price = (float) $request->data('price');
-$isActive = (bool) $request['is_active'];
+$name = $request->string('name');
+$age = $request->integer('age');
+$price = $request->float('price');
+$isActive = $request->boolean('is_active');
```

### DIFF
--- a/tests/Rector/MethodCall/RequestInputToTypedMethodRector/Fixture/array_access_with_cast.php.inc
+++ b/tests/Rector/MethodCall/RequestInputToTypedMethodRector/Fixture/array_access_with_cast.php.inc
@@ -1,0 +1,37 @@
+<?php
+
+namespace RectorLaravel\Tests\Rector\MethodCall\RequestInputToTypedMethodRector\Fixture;
+
+use Illuminate\Http\Request;
+
+class ArrayAccessWithCast
+{
+    public function store(Request $request)
+    {
+        $name = (string) $request['name'];
+        $age = (int) $request['age'];
+        $price = (float) $request['price'];
+        $isActive = (bool) $request['is_active'];
+    }
+}
+
+?>
+-----
+<?php
+
+namespace RectorLaravel\Tests\Rector\MethodCall\RequestInputToTypedMethodRector\Fixture;
+
+use Illuminate\Http\Request;
+
+class ArrayAccessWithCast
+{
+    public function store(Request $request)
+    {
+        $name = $request->string('name');
+        $age = $request->integer('age');
+        $price = $request->float('price');
+        $isActive = $request->boolean('is_active');
+    }
+}
+
+?>
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1,4 +1,4 @@
-# 86 Rules Overview
+# 87 Rules Overview
 
 ## AbortIfRector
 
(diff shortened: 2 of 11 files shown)


alexey-m-ukolov:
I like the idea just want to point out that `$request->input('name')` and `$request->string('name')` are not the same - former returns (in this example) `string|null` and latter returns `\Illuminate\Support\Str`.
I'm not sure if this case should be handled at all - even if we replace with `$request->string('name')->toString()` we are still losing `null`. I think only the explicit casts and calls like `$request->input('name', '')` can be safely replaced (with added `->toString()`).

tests/Rector/MethodCall/RequestInputToTypedMethodRector/Fixture/array_access_with_cast.php.inc
```diff
@@ -0,0 +1,37 @@
...
+use Illuminate\Http\Request;
+
+class ArrayAccessWithCast
+{
+    public function store(Request $request)
+    {
+        $name = (string) $request['name'];
+        $age = (int) $request['age'];
+        $price = (float) $request['price'];
+        $isActive = (bool) $request['is_active'];
```
peterfox: Might want to consider what happens if null coalese is used:

```suggestion
        $name = (string) $request['name'];
        $age = (int) $request['age'];
        $price = (float) $request['price'];
        $isActive = (bool) $request['is_active'];
        $name = (string) ($request['name'] ?? null);
        $age = (int) ($request['age'] ?? null);
        $price = (float) ($request['price'] ?? null);
        $isActive = (bool) ($request['is_active'] ?? null);
```

yoannblot:
We've been running a rule for this in our codebase (Laravel 13, a few hundred controllers) for a while. I'd be happy to open a PR for it. Here is how it handles the concern @alexey-m-ukolov raised: it only rewrites when the result is provably the same.

**1. Explicit cast**: the cast already throws away `null`, so the typed accessor returns the same value:

```diff
-$isEdito  = (bool) $request->get('is_edito', false);
-$autoplay = (bool) $request->get('autoplay', true);
-$perPage  = (int) $request->get('per_page', self::PER_PAGE);
-$volume   = (float) $request->get('volume', 0);
-$query    = (string) $request->get('query', '');
-$terms    = explode(',', (string) $request->get('terms'));
+$isEdito  = $request->boolean('is_edito');
+$autoplay = $request->boolean('autoplay', true);
+$perPage  = $request->integer('per_page', self::PER_PAGE);
+$volume   = $request->float('volume');
+$query    = $request->string('query')->toString();
+$terms    = explode(',', $request->string('terms')->toString());
```

- A default that equals the accessor's own default (`false`, `0`, `''`) is dropped.
- `string()` always gets `->toString()`, so the result is still a `string`, never a `Stringable`.
- A cast on a value that is already typed (`(int) $request->integer('page')`) is removed.

**2. `[]` default, or the value used as an array**:

```diff
-$metrics  = $request->get('metrics', []);
-$objectId = $request->input('object')['id'];
+$metrics  = $request->array('metrics');
+$objectId = $request->array('object')['id'];
```

**3. No cast, type taken from the FormRequest** (optional second step): when the parameter is a
custom `FormRequest` and its `rules()` returns a literal array that declares the field as `required` with
a type rule (`bool`, `int`, `float`, `array`; the upstream version would also accept `boolean`/`integer`), the value cannot be `null`
once validation has passed. `string` is left out on purpose, because `string()` returns a `Stringable`:

```diff
 // rules(): 'is_active' => ['required', 'bool']
-$isActive = $request->input('is_active');
+$isActive = $request->boolean('is_active');
```

**Skipped on purpose** (each has a fixture):
- a non-literal key, or a dotted/`*` key: `get()` reads the flat key, while `input()` and the typed
  accessors go through `data_get()`;
- `$request->query->get()`, `->headers->get()`, `->request->get()`: these are Symfony bags, not the input;
- a plain `Symfony\Component\HttpFoundation\Request`, or any receiver that is not an Illuminate `Request`;
- a non-empty array default (`(array) $request->get('ids', [1])`);
- for step 3: a field that is not `required`, a field without a type rule, and `rules()` that is not a literal array.

My suggestion: a first PR with 1 + 2 only, since they are purely local and have no false positives. Step 3
would come as a follow-up, or behind a config option, since it needs to read the FormRequest. Would that work
for you?

PR #416, opened 2025-10-31
Bump Rector requirement to ^2.2.7

samsonasik:
@Arcont01 @bitw @TomasVotruba this can possibly a way to handle rector reported issue:

- https://github.com/rectorphp/rector/issues/9470

ensure it uses latest rector 2.2.7+ that already include latest nikic/php-parser inside scoped vendor when `getArg()` exists

https://github.com/rectorphp/rector/blob/022038537838bc8a4e526af86c2d6e38eaeff7ef/vendor/nikic/php-parser/lib/PhpParser/Node/Expr/CallLike.php#L43


samsonasik:

PR #422, opened 2025-11-09
fix: deficiencies in the EloquentMagicMethodToQueryBuilderRector

calebdw:
Hello!

This fixes several deficiencies in the rector that prevented the following cases from being converted:
- static calls on class-strings
- methods on custom builder classes
- calls on an unknown models
- calls on a model intersection with an interface
- static calls with a model scope

Thanks!


peterfox:
This looks really useful @calebdw thank you for your contribution!

PR #424, opened 2025-11-20
Ensure reset key of stmts/args before returning node when there is unset by key

samsonasik:
Rector is possibly removing auto indexing key soon, ref for this starter:

- https://github.com/rectorphp/rector-src/pull/7643

and follow up PRs on latest main branch. This ensure reindex key when `unset($node->args[key])` node data, eg: args and stmts.


samsonasik:

PR #426, opened 2025-11-21
fix: enhance ArrayToArrGetRector to skip isset, empty checks, and assignments

### LINKED ISSUE #425: `ArrayToArrGetRector` fails when array fetch is in `isset` or `empty`
it converts `isset($arr['key'])` to `isset(Arr::get($arr, 'key'))` which is not good

MrPunyapal:
fixes #425

### DIFF
--- a/src/Rector/ArrayDimFetch/ArrayToArrGetRector.php
+++ b/src/Rector/ArrayDimFetch/ArrayToArrGetRector.php
@@ -106,7 +106,6 @@ CODE_SAMPLE
     }
 
     /**
-     * @param Coalesce $coalesce
      * @return StaticCall|NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN
      */
     private function refactorCoalesce(Coalesce $coalesce): StaticCall|int


MrPunyapal:
@peterfox

peterfox:
There's actually a better way to handle this by telling rector to ignore further nodes but I'll look into that.

You'll also want to look at uses of unset as that'll be the same as empty/isset

MrPunyapal:
@peterfox I have added the `unset` and I would love to learn about how we can improve it

peterfox:
@MrPunyapal I've worked on an offshoot of your branch. If you want to pull these changes into yours and go from there: https://github.com/driftingly/rector-laravel/compare/fix/clean-up-ArrayToArrGetRector

What I've changed is I've simplified your rule so it's no longer storing the references to other nodes. Instead I make use of the PHP Parser's NodeVisitor constants. By returning `NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN` you can tell the parser to skip going further down the tree.

This means things like Isset and Unset can be found earlier and then be ignored if a ArrayDimFetch is found.

I've also implemented a slightly nicer way to look into the further nodes with this method:

```php
private function containsArrayDimFetch(Node $node): bool
    {
        $found = false;

        $this->traverseNodesWithCallable($node, function (Node $node) use (&$found): ?int {
            if ($node instanceof ArrayDimFetch) {
                $found = true;

                return NodeVisitor::STOP_TRAVERSAL;
            }

            return null;
        });

        return $found;
    }
```

Using `traverseNodesWithCallable()` allows us to look through further nodes in the same way Rector uses rules to do so. This makes it easier than having to examine each Node type and determine what properties might contain the ArrayDimFetch we're looking for. We return `NodeVisitor::STOP_TRAVERSAL` to terminate the search vertically and horizontally across the nodes as we're already found one.

MrPunyapal:
@peterfox done

PR #429, opened 2025-11-27
feat: rename resolve() to app()

calebdw:
Hello!

`resolve()` is just an alias of `app()`, so this renames all instances to `app()` which also slightly increases performance

Thanks!

### DIFF
--- a/config/sets/laravel-code-quality.php
+++ b/config/sets/laravel-code-quality.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 use Rector\Config\RectorConfig;
+use Rector\Renaming\Rector\FuncCall\RenameFunctionRector;
 use RectorLaravel\Rector\ArrayDimFetch\EnvVariableToEnvHelperRector;
 use RectorLaravel\Rector\ArrayDimFetch\RequestVariablesToRequestFacadeRector;
 use RectorLaravel\Rector\ArrayDimFetch\ServerVariableToRequestFacadeRector;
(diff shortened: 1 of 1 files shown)


peterfox:
I'm not sure I agree on this being a code quality choice. In part because people might use `resolve()` for better IDE type hinting while `app()` does additional magic. If anything I'd more be inclinded to make a rule that determines when it's better to use `resolve()` over `app()` when applicable.

calebdw:
`app()` has the same function phpdocs that `resolve()` does so it will work equally as well. The only additional thing that `app()` does is return the container when there's no arguments provided which is also handled by the phpdocs.

What heuristics would you use to determine when to rename? I can't think of any... Also, there's other rules that replaced aliased methods with the actual method, so I don't really see this as any different

peterfox:
> What heuristics would you use to determine when to rename? I can't think of any... Also, there's other rules that replaced aliased methods with the actual method, so I don't really see this as any different

Can you supply some examples? Note ones where the soft signature of the methods aren't the same.

> app() has the same function phpdocs that resolve() does so it will work equally as well. The only additional thing that app() does is return the container when there's no arguments provided which is also handled by the phpdocs.

While it does use `app()` the soft signatures are quite different:

```php
    /**
     * Resolve a service from the container.
     *
     * @template TClass of object
     *
     * @param  string|class-string<TClass>  $name
     * @return ($name is class-string<TClass> ? TClass : mixed)
     */
    function resolve($name, array $parameters = [])
```

```php
    /**
     * Get the available container instance.
     *
     * @template TClass of object
     *
     * @param  string|class-string<TClass>|null  $abstract
     * @return ($abstract is class-string<TClass> ? TClass : ($abstract is null ? \Illuminate\Foundation\Application : mixed))
     */
    function app($abstract = null, array $parameters = [])
    {
        if (is_null($abstract)) {
            return Container::getInstance();
        }

        return Container::getInstance()->make($abstract, $parameters);
    }
```

I personally would use `resolve` over `app` more often than not.

e.g.
```
resolve(Foo::class)
// is a lot clearer than
app(Foo::class)
// because app() is more ambiguous, if I use app() to resolve something I usually would use:
app()->make(Foo::class) 
```

calebdw:
Hmm,  I can write a rule to go from `app(Foo::class)` to `resolve(Foo::class)`, there should be some consistency one way or another

peterfox:
I agree and that would be a useful rule to have.

PR #430, opened 2025-11-29
feat: add app to resolve rector

calebdw:
Supersedes #429 

@peterfox, here you go!


peterfox:
Looks great. Thank you @calebdw ! 🎉

ManuelLeiner:
Hey guys,
Thanks for the new rule. In general it makes sense to replace `app` by `resolve`.. it is more precise. Unfortunately, it seems to be broken: #432.

PR #431, opened 2025-11-29
fix: bugs in eloquent magic method to query builder rector

calebdw:
@peterfox, this fixes two bugs:

- potential for non model class reflection to persist after the loop
- need to use `hasNativeMethod` to keep Larastan mixins from detecting the method on the model


peterfox:
Thank you again @calebdw 🎉

PR #433, opened 2025-12-01
fix: app to resolve rector

### LINKED ISSUE #432: [BUG] `AppToResolveRector` breaks invokable classes used as Laravel action objects
Since version **2.1.4**, the Rector rule `AppToResolveRector` was introduced. This rule does not handle **invokable PHP classes** correctly, especially when following Laravel’s common *action pattern*.

## Problematic transformation

```diff
- app(EnableTwoFactorAuthentication::class)($user);
+ resolve($user);

------------ end diff ------------
Applied rules:
 * AppToResolveRector
```

The transformed code breaks the application. `$user` is passed to `resolve()` instead of the action class, so the invokable class is not instantiated.

Of course, it works if I create an action variable first.

```diff
-    $action = app(EnableTwoFactorAuthentication::class);
+    $action = resolve(EnableTwoFactorAuthentication::class);
     $action($user);
```
## Expected transformation

```diff
- app(EnableTwoFactorAuthentication::class)($user);
+ resolve(EnableTwoFactorAuthentication::class)($user);

------------ end diff ------------
Applied rules:
 * AppToResolveRector
```

calebdw:
Closes #432 

Thanks!

CC: @manuelLeiner, @peterfox

### DIFF
--- a/src/Rector/FuncCall/AppToResolveRector.php
+++ b/src/Rector/FuncCall/AppToResolveRector.php
@@ -42,7 +42,10 @@ CODE_SAMPLE
      */
     public function refactor(Node $node): ?FuncCall
     {
-        if (! $this->isName($node->name, 'app')) {
+        if (
+            ! $node->name instanceof Name
+            || ! $this->isName($node->name, 'app')
+        ) {
             return null;
         }
 
--- a/tests/Rector/FuncCall/AppToResolveRector/Fixture/fixture.php.inc
+++ b/tests/Rector/FuncCall/AppToResolveRector/Fixture/fixture.php.inc
@@ -4,6 +4,9 @@ namespace RectorLaravel\Tests\Rector\FuncCall\AppToResolveRector\Fixture;
 
 app('foo');
 app('bar', ['foo' => 'bar']);
+app('baz')->foo();
+app('baq')();
+app('baq')(5);
 
 ?>
 -----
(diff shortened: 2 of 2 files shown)


ManuelLeiner:
Works for me!

37 files have been changed by Rector and all 227 tests pass. Diff looks as expected.

calebdw:
@samsonasik, could you provide some insight here? It looks like `$this->isName($node->name, 'app')` returns true for a nested function call (which should not be the case?)---might be a bug in the `isName` method?

```php
app('foo')($bar);
```

This is a nested function call, but the outer `name` (`app('foo')`) should not return `true` for `$this->isName($node->name, 'app')`?

```
PhpParser\Node\Stmt\Expression(
    expr: PhpParser\Node\Expr\FuncCall(
        name: PhpParser\Node\Expr\FuncCall(
            name: PhpParser\Node\Name\FullyQualified( parts: ["app"] )
            args: [
                0: PhpParser\Node\Arg(
                    name: null
                    value: PhpParser\Node\Scalar\String_( value: "foo" )
                    byRef: false
                    unpack: false
                )
            ]
        )
        args: [
            0: PhpParser\Node\Arg(
                name: null
                value: PhpParser\Node\Expr\Variable( name: "bar" )
                byRef: false
                unpack: false
            )
        ]
    )
)
```

samsonasik:
The usage for function call should be just:

```php
if (! $this->isName($node, 'app')) {
```

PR #435, opened 2025-12-02
Reduce NodeVisitor::DONT_TRAVERSE_* usage on ArrayToArrGetRector (part 1)

samsonasik:
Rector is removing tweak below node usage via:

```
NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN
NodeVisitor::DONT_TRAVERSE_CHILDREN
```

see PR:

- https://github.com/rectorphp/rector-src/pull/7697

This PR is starter to keep compatible :)

### DIFF
--- a/src/Rector/ArrayDimFetch/ArrayToArrGetRector.php
+++ b/src/Rector/ArrayDimFetch/ArrayToArrGetRector.php
@@ -94,7 +94,7 @@ CODE_SAMPLE
             return null;
         }
 
-        if ($node->getAttribute(AttributeKey::IS_BEING_ASSIGNED)) {
+        if ($node->getAttribute(AttributeKey::IS_BEING_ASSIGNED) === true) {
             return null;
         }
 


samsonasik:

It will take multiple PRs as there are many usage of it in this repo, this is for starter :)

peterfox:
Is the idea to now use NodeVisitors to mark nodes instead of stopping transversal from what I can see in that PR you mentioned @samsonasik ?

samsonasik:
@peterfox I am thinking of set flag attribute via `beforeTraverse()` on each rule that needs it, I will create next PR :)

samsonasik:
@peterfox the stop traversal only works on single visitor (eg: callable traverser).

On rector rules, it will stop immediatelly the loop without next rule, so the attribute flag is the way to go.

peterfox:
> @peterfox I am thinking of set flag attribute via `beforeTraverse()` on each rule that needs it, I will create next PR :)

Okay, do you have any examples I can look at?

samsonasik:
I on mobile, will create PR here today, for reuse here, custom node visitor decorate maybe needed.

samsonasik:
@peterfox here usage on `beforeTraverse()` :)

- https://github.com/driftingly/rector-laravel/pull/437

peterfox:
@samsonasik great. So essentially all nodes can be examined with `beforeTraverse`, then add the attributes to node which might be problematic etc. and need skipping.

PR #436, opened 2025-12-02
docs: improve rule docs and readme

peaklabs-dev:
## Changes

- added **Sets** to each rule which makes it a lot easier to check if a rule is applied already or needs to be added manually
- added missing new rule `AppToResolveRector`
- improve `MinutesToSecondsInCacheRector` rule description
- improve multiple README sections for better understandability
- rewrote the manual configuration section for easier understanding
- added some details to sets that are helpful for deciding to add all rules manually or to use sets (some options are only available via sets for example)
- added missing configurable and opinionated rules to README file
- sorted opinionated rules alphabetically

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1,4 +1,4 @@
-# 88 Rules Overview
+# Rules Overview
 
 ## AbortIfRector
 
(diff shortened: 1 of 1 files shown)


peterfox:
I think this was a good idea for the README.md but the rules document is actually autogenerated by the `composer docs` command. If anything we probably should do the same with sets to update the README.md file or generate a new one just for sets.

PR #438, opened 2025-12-03
Additional Analysers for Models/Query Builder

peterfox:
This adds:

* A way to analyse if a method on a model is in fact a Query Scope method used in Eloquent queries.
* A way to analyse if a Eloquent query builder is for a specified model.
* A way to analyse if a method on a model is a relation and what Related and Parent models are used in the relation.
* Covers tests for the new features.

Improves:

* Some PHPDocs
* What exceptions are thrown
* Namespace issues in some tests
* Allows strings or ObjectTypes to be passed to a number of previously made analyser methods.

### DIFF
--- a/src/NodeAnalyzer/ModelAnalyzer.php
+++ b/src/NodeAnalyzer/ModelAnalyzer.php
@@ -7,7 +7,6 @@ use InvalidArgumentException;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\ExtendedMethodReflection;
-use PHPStan\Reflection\MissingMethodFromReflectionException;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\ObjectType;
 use ReflectionException;
@@ -88,16 +87,13 @@ class ModelAnalyzer
             return true;
         }
 
-        try {
-            $method = $classReflection->getMethod($scopeName, $scope);
-
-            if ($this->usesScopeAttribute($method)) {
-                return true;
-            }
-        } catch (MissingMethodFromReflectionException) {
+        if (! $classReflection->hasMethod($scopeName)) {
+            return false;
         }
 
-        return false;
+        $extendedMethodReflection = $classReflection->getMethod($scopeName, $scope);
+
+        return $this->usesScopeAttribute($extendedMethodReflection);
     }
 
     /**
--- a/src/NodeAnalyzer/QueryBuilderAnalyzer.php
+++ b/src/NodeAnalyzer/QueryBuilderAnalyzer.php
@@ -97,7 +97,7 @@ final readonly class QueryBuilderAnalyzer
      */
     public function resolveQueryBuilderModel(Type $objectType, Scope $scope): ?Type
     {
-        if ($objectType->isObject()->no() || $objectType->isSuperTypeOf(self::queryBuilderType())->no()) {
+        if (self::queryBuilderType()->isSuperTypeOf($objectType)->no()) {
             throw new InvalidArgumentException('Object type must be an Eloquent query builder.');
         }
 


src/NodeAnalyzer/ModelAnalyzer.php
```diff
@@ -52,26 +54,89 @@ public function getPrimaryKey(string $class): ?string
...
+     * @return Model
+     * @throws ReflectionException
+     */
+    private function resolveModelClassToInstance(string|ObjectType $model): Model
+    {
+        if (! is_string($model)) {
+            $model = $model->getClassName();
+        }
+
+        $classReflection = $this->getClass($model);
```
calebdw: There's no need to get the class name from the object type and then get the class reflection when you can directly get the class reflection:


```suggestion
        $classReflection = is_string($model)
             ? $this->getClass($model)
             : $model->getObjectClassReflections()[0];
```

src/NodeAnalyzer/QueryBuilderAnalyzer.php
```diff
@@ -26,6 +31,19 @@ protected static function queryBuilderType(): ObjectType
         return new ObjectType('Illuminate\Contracts\Database\Query\Builder');
     }
 
+    protected static function eloquentQueryBuilderType(): ObjectType
+    {
+        return new ObjectType('Illuminate\Contracts\Database\Eloquent\Builder');
```
calebdw: This should use the actual builder, not the contract (I hate the contracts: they are empty and not true contracts, and the Relation also implements the contract which is not great)


```suggestion
        return new ObjectType('Illuminate\Database\Eloquent\Builder');
```

peterfox: Fair. My slight worry is that most will type hint the contract still and it might then cause the rule to miss things. In the analyser it should be fine.

peterfox:
@calebdw I'll leave this PR here. I've added some relation analysis methods but I don't want to go overboard with creating lots of cool tooling without them being used more first.

src/NodeAnalyzer/QueryBuilderAnalyzer.php
```diff
@@ -66,4 +87,50 @@ public function isProxyCall(StaticCall $staticCall): bool
...
+    public function resolveQueryBuilderModel(Type $objectType, Scope $scope): ?Type
+    {
+        if ($objectType->isObject()->no() || $objectType->isSuperTypeOf(self::queryBuilderType())->no()) {
+            throw new InvalidArgumentException('Object type must be an Eloquent query builder.');
+        }
+
+        $extendedPropertyReflection = $objectType->getInstanceProperty('model', $scope);
+        $modelType = $extendedPropertyReflection->getReadableType();
+
+        if ($modelType->isSuperTypeOf(self::modelType())->no()) {
```
calebdw:
```suggestion
        if (self::modelType()->isSuperTypeOf($modelType)->no()) {
```

src/NodeAnalyzer/ModelAnalyzer.php
```diff
@@ -52,26 +73,122 @@ public function getPrimaryKey(string $class): ?string
...
+
+        if (! $classReflection->hasMethod($relationName)) {
+            return false;
+        }
+
+        $extendedMethodReflection = $classReflection->getMethod($relationName, $scope);
+
+        foreach ($extendedMethodReflection->getVariants() as $extendedParametersAcceptor) {
+            if ($extendedParametersAcceptor->getReturnType()->isSuperTypeOf(self::relationType())->yes()) {
```
calebdw:
```suggestion
            if (self::relationType()->isSuperTypeOf($extendedParametersAcceptor->getReturnType())->yes()) {
```

src/NodeAnalyzer/RelationshipAnalyzer.php
```diff
@@ -0,0 +1,65 @@
...
+        $extendedPropertyReflection = $objectType->getInstanceProperty('related', $scope);
+        $modelType = $extendedPropertyReflection->getReadableType();
+
+        if ($modelType->isSuperTypeOf(self::modelType())->no()) {
+            return null;
+        }
+
+        /** @phpstan-ignore return.type */
+        return $modelType;
+    }
```
calebdw: - same as before, should be getting the inner type from the templates, not a protected property
- no need to throw an exception because `getTemplateType` is on the Type api

```suggestion
    public function resolveRelatedForRelation(Type $objectType): ?Type
    {
        $modelType = $objectType->getTemplateType(Relation::class, 'TRelatedModel');

        if (self::modelType()->isSuperTypeOf($modelType)->no()) {
            return null;
        }

        /** @phpstan-ignore return.type */
        return $modelType;
    }
```

src/NodeAnalyzer/RelationshipAnalyzer.php
```diff
@@ -0,0 +1,65 @@
...
+        $extendedPropertyReflection = $objectType->getInstanceProperty('parent', $scope);
+        $modelType = $extendedPropertyReflection->getReadableType();
+
+        if ($modelType->isSuperTypeOf(self::modelType())->no()) {
+            return null;
+        }
+
+        /** @phpstan-ignore return.type */
+        return $modelType;
+    }
```
calebdw: I doubt this will ever be needed (I've never needed the parent in Larastan), but here's the correct version


```suggestion
    /**
     * Resolve the Parent Model of the Relationship.
     *
     * @return ObjectType|null
     */
    public function resolveParentForRelation(Type $objectType): ?Type
    {
        $modelType = $objectType->getTemplateType(Relation::class, 'TDeclaringModel');
        
        if (self::modelType()->isSuperTypeOf($modelType)->no()) {
            return null;
        }
        
        /** @phpstan-ignore return.type */
        return $modelType;
    }
```

src/NodeAnalyzer/ModelAnalyzer.php
peterfox: Are you sure for this one? When I change it, the tests break for scenario `it_can_determine_if_the_model_has_relationship_method` at line `tests/NodeAnalyzer/ModelAnalyzerTest.php:117`

peterfox: Would PHPStan now infer the correct type purely from the fact it always returns a HasMany object anyways?

peterfox: Well you're right although I'm a little confused still because I assumed it would infer the type from the PHPDocs and the return of the method but I guess not.

calebdw: PHPStan doesn't descend into method and functions---it stops at the boundary. When inside a method phpstan checks that you return what you say you return, and when outside a method phpstan checks to make sure the arguments you pass are compatible 

https://phpstan.org/user-guide/troubleshooting-types#what-phpstan-doesn%E2%80%99t-do

peterfox: Ah okay. That makes sense. Thank you for your patience with me.

src/NodeAnalyzer/ModelAnalyzer.php
```diff
@@ -52,26 +73,128 @@ public function getPrimaryKey(string $class): ?string
...
+            $method = $classReflection->getMethod($scopeName, $scope);
+
+            if ($this->usesScopeAttribute($method)) {
+                return true;
+            }
+        } catch (MissingMethodFromReflectionException) {
+        }
+
+        return false;
+    }
```
calebdw: nit: you can remove the need for a try/catch by checking if the class has the method first:


```suggestion
        if (! $classReflection->hasMethod($scopeName)) {
            return false;
        }

        $method = $classReflection->getMethod($scopeName, $scope);

        return $this->usesScopeAttribute($method);
```

src/NodeAnalyzer/QueryBuilderAnalyzer.php
```diff
@@ -66,4 +87,50 @@ public function isProxyCall(StaticCall $staticCall): bool
...
+     *
+     * @return ObjectType|null
+     *
+     * @throws \PHPStan\ShouldNotHappenException
+     */
+    public function resolveQueryBuilderModel(Type $objectType, Scope $scope): ?Type
+    {
+        if ($objectType->isObject()->no() || $objectType->isSuperTypeOf(self::queryBuilderType())->no()) {
+            throw new InvalidArgumentException('Object type must be an Eloquent query builder.');
+        }
```
calebdw:
```suggestion
        if (self::queryBuilderType()->isSuperTypeOf($objectType)->no()) {
            throw new InvalidArgumentException('Object type must be an Eloquent query builder.');
        }
```

src/NodeAnalyzer/ModelAnalyzer.php
calebdw: No worries! It's a common misconception, but it helps to better understand how phpstan works once you grasp it :smiley:

PR #441, opened 2025-12-03
Remove NodeVisitor::DONT_TRAVERSE_* on SessionVariableToSessionFacadeRector (part 5)

samsonasik:
Continue of PR:

- https://github.com/driftingly/rector-laravel/pull/440

PR #442, opened 2025-12-04
Remove return NodeVisitor::DONT_TRAVERSE_* on ArrayToArrGetRector (part 6)

samsonasik:
Continue of PR:

- https://github.com/driftingly/rector-laravel/pull/441


samsonasik:
@GeniJaho @driftingly @peterfox Ready to merge 👍

This is the last clean up of `NodeVisitor::DONT_TRAVERSE_*`, after this PR merged, I think new immediate release is needed so next composer update when sometime rector released new version, there will be less to no regression as possible :)

peterfox:
@samsonasik should we set rector to a certain minimum version if it'll cause breakages?

samsonasik:
This change is make code compatible, so no breaking change here when next rector sometime released, updating to rector/rector:^2.2.11 still fine here imo

PR #443, opened 2025-12-07
Move beforeTraverse() to global Node since beforeTraverse() not overrideable

samsonasik:
since `beforeTraverse()` become final, 

- https://github.com/rectorphp/rector-src/pull/7727

the use of beforeTraverse here is invalid,

Using custom node visitor is not ideal for shareable rules, since user then need another step to register visitor, this PR make it keep working automatically, by register on global `Node` early on the particular rules that requires set attributes.

### DIFF
--- a/src/NodeAnalyzer/ExpectedClassMethodAnalyzer.php
+++ b/src/NodeAnalyzer/ExpectedClassMethodAnalyzer.php
@@ -18,7 +18,6 @@ use RectorLaravel\ValueObject\ExpectedClassMethodMethodCalls;
 final readonly class ExpectedClassMethodAnalyzer
 {
     public function __construct(
-        private SimpleCallableNodeTraverser $simpleCallableNodeTraverser,
         private NodeNameResolver $nodeNameResolver,
         private NodeTypeResolver $nodeTypeResolver,
     ) {}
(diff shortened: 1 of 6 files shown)


samsonasik:
@GeniJaho @driftingly @peterfox Ready to merge 👍

after this PR merged, I think new immediate release is needed so next composer update when sometime rector released new version, there will be less to no regression as possible :)

samsonasik:
For note: after next rector release, which new `FileNode` will exists, instead of global node that can be any node, the `FileNode` and `Namespace_` should be used instead as a "top" node :)

This PR ensure keep work first whenever next rector release anytime soon when we are may not available to immediatelly patch :)

samsonasik:
@GeniJaho @driftingly @peterfox I've bumped to latest rector ^2.2.13 and fix compatible code:

- ensure use `isInFirstLevelStatement()` from scope to reduce repetitive init traverse https://github.com/driftingly/rector-laravel/pull/443/commits/9cc30cf4a29be848a4efb49172b68fa8a999c15e 
- fix patch of static `SimpleCallableNodeTraverser::traverseNodesWithCallable` usage  https://github.com/driftingly/rector-laravel/pull/443/commits/c1e50cc11e11db510fefd0c74a4c9e23f9b836cc 

Ready to merge and this needs immediate new release :)

peterfox:
Looks good. I'm wandering if what we need is a mechanism where a default config is loaded with the Laravel Package or the package has it's own Config object at this point that preloads a few things. Then we could use NodeVisitors.

This will be good to resolve the current situation.

samsonasik:
There is default config:

https://github.com/driftingly/rector-laravel/blob/main/config/config.php

already included in sets, but then user that only want to use single rule and done, will require manually include it, that's another step for user.

PR #444, opened 2025-12-10
[alternative] Use NodeVisitor for init attribute flag over global Node traverse

samsonasik:
@peterfox per your feedback at https://github.com/driftingly/rector-laravel/pull/443#issuecomment-3632859202

This move to use `NodeVisitor` over global Node. Here the benefit and drawback:

**Benefit**:

- can reuse class constant attribute over rules if needed.

**Drawback**:

- The default config is included in sets, but if user just want to use some rule that require the attribute, user need to import the default config:

```php
->withSets([
     __DIR__ . '/vendor/driftingly/rector-laravel/config/config.php'
])
->withRules([
     // the rule from this repo :)
]);
```

I've updated readme for it tho.

### DIFF
--- a/.github/workflows/code_analysis.yaml
+++ b/.github/workflows/code_analysis.yaml
@@ -36,13 +36,13 @@ jobs:
         name: ${{ matrix.actions.name }}
 
         steps:
-            -   uses: actions/checkout@v4
+            -   uses: actions/checkout@v6
 
             -   uses: shivammathur/setup-php@v2
                 with:
                     php-version: 8.3
                     coverage: none
 
-            -   uses: ramsey/composer-install@v2
+            -   uses: ramsey/composer-install@v3
 
             -   run: ${{ matrix.actions.run }}
--- a/.github/workflows/downgraded_release.yaml
+++ b/.github/workflows/downgraded_release.yaml
@@ -15,7 +15,7 @@ jobs:
 
         steps:
             -
-                uses: "actions/checkout@v4"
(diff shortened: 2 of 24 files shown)


samsonasik:
@GeniJaho @driftingly @peterfox ready for review/merge 👍

calebdw:
@samsonasik, is there any way for this package to automatically register this config so users don't have to when using single rules?

samsonasik:
@calebdw I am on mobile, not sure, there is rector-extension-installer, but currently only used under rector organization that require-dev of rector-src include usage and require php 8

https://github.com/rectorphp/extension-installer/tree/main?tab=readme-ov-file#instructions-for-extension-developers

calebdw:
Ah, looks handy! Do you think we could use it as part of this PR?

samsonasik:
I am not sure as that require php 8.0, and this repo require php 7.4

https://github.com/driftingly/rector-laravel/blob/aee9d4a1d489e7ec484fc79f33137f8ee051b3f7/composer.json#L7

On rector, the generated config updated on build, while on this, this means that needs to be part of "require"

calebdw:
I don't think this repo needs to `require` the `rector-extension-installer`, we just have to include the `rector` key in the `composer.json`

For example, phpstan packages don't need to require the `phpstan-extension-installer`, just include the config in the composer.json

samsonasik:
I currently don't have much energy (less sleep due to PHPStan 2.1.34 yesterday), could you handle that ? Thank you.

PR #449, opened 2025-12-14
Stricter checks for converting startOfDay() to today()

### LINKED ISSUE #448: Potential for incorrect fixes with NowFuncWithStartOfDayMethodCallToTodayFuncRector
The [NowFuncWithStartOfDayMethodCallToTodayFuncRector](https://github.com/driftingly/rector-laravel/blob/main/src/Rector/FuncCall/NowFuncWithStartOfDayMethodCallToTodayFuncRector.php) rule seems a little bit too eager to swap out calls to `startOfDay()` with `today()`

For example with the following example class:

```php
<?php

use Carbon\CarbonImmutable;

class Example
{
    private CarbonImmutable $now;

    public function __construct(?CarbonImmutable $now = null)
    {
        $this->now = $now ?? now()->toImmutable();
    }

    private function getStartDate(): CarbonImmutable
    {
        return $this->now->startOfDay();
    }
}
```

This rule will incorrectly convert the call `startOfDay()` to `today()` and break the code

```diff
<?php

use Carbon\CarbonImmutable;

class Example
{
    private CarbonImmutable $now;

    public function __construct(?CarbonImmutable $now = null)
    {
        $this->now = $now ?? now()->toImmutable();
    }

    private function getStartDate(): CarbonImmutable
    {
-        return $this->now->startOfDay();
+        return today();
    }
}
```

It seems like anything called "now" on the left hand side will get converted, rather than just `now()` helper calls.

philo23:
Fixes #448 where this rule will incorrectly match any variable on the left hand side of the call to the startOfDay() method as long as its called "now" rather than just the now() helper.

Eg: $now->startOfDay(), $this->now->startOfDay() or $this->now()->startOfDay()

### DIFF
--- a/src/Rector/FuncCall/NowFuncWithStartOfDayMethodCallToTodayFuncRector.php
+++ b/src/Rector/FuncCall/NowFuncWithStartOfDayMethodCallToTodayFuncRector.php
@@ -45,6 +45,10 @@ CODE_SAMPLE
             return null;
         }
 
+        if (! $node->var instanceof FuncCall) {
+            return null;
+        }
+
         if (! $this->isName($node->var, 'now')) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


peterfox:
Thank you for this fix @philo23, good catch!

PR #450, opened 2025-12-16
[Feature] Where comparison to null refactored whereNull rule

peterfox:
# Why

It makes sense to not use `where('foo', '=', null)` and instead use `whereNull('foo')`.

# Example

```diff
-$query->where('foo', null);
-$query->where('foo', '=', null);
-$query->where('foo');
-$query->where('foo', '!=', null);
+$query->whereNull('foo');
+$query->whereNull('foo');
+$query->whereNull('foo');
+$query->whereNotNull('foo');
```

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1806,9 +1806,11 @@ Convert to where comparison to whereNull method call
 -$query->where('foo', null);
 -$query->where('foo', '=', null);
 -$query->where('foo');
+-$query->where('foo', '!=', null);
 +$query->whereNull('foo');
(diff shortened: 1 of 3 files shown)


MrPunyapal:
can it also have `whereNotNull` too?

```diff
-where('column', '!=', null)
+whereNotNull('column')
```

peterfox:
@MrPunyapal done!

pindab0ter:
Does this support variables which you know to be `null`?

peterfox:
It would yes. If the code is certain that only null can be provided

PR #458, opened 2026-01-04
[2.x] Update action versions

jackbayliss:
Noticed some outdated actions here while I was about to PR a new rule, so thought I'd sort it.

- [Checkout is v6](https://github.com/actions/checkout?tab=readme-ov-file#checkout-v6)
- [composer-install is v3](https://github.com/ramsey/composer-install/releases)


peterfox (approved):

peterfox:
Thank you @jackbayliss , appreciate this

calebdw:
we should just enable dependabot

PR #459, opened 2026-01-05
Adds AddReturnTypeToModelRelationshipMethodRector rule

peterfox:
# Changes

New rule AddReturnTypeToModelRelationshipMethodRector
Adds tests
Updates docs
Added the rule to the type improvements set

# Why

PHPStan can't detect relationships unless the return type is there so having a way to automate adding those types makes sense.

```diff
 class User extends Model
 {
-    public function posts()
+    public function posts(): \Illuminate\Database\Eloquent\Relations\HasMany
     {
         return $this->hasMany(Post::class);
     }
 }
```

### DIFF
--- a/.github/workflows/code_analysis.yaml
+++ b/.github/workflows/code_analysis.yaml
@@ -36,13 +36,13 @@ jobs:
         name: ${{ matrix.actions.name }}
 
         steps:
-            -   uses: actions/checkout@v4
+            -   uses: actions/checkout@v6
 
             -   uses: shivammathur/setup-php@v2
                 with:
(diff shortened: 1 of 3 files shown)


calebdw:
This shouldn't be needed, there should already be rector rules that adds simple return types like this for returned method calls. after the return type has been added then https://github.com/driftingly/rector-laravel/blob/main/src/Rector/ClassMethod/AddGenericReturnTypeToRelationsRector.php can update the generics

peterfox:
Yep, I should have looked harder. I assumed we didn't have it

PR #463, opened 2026-01-17
New rule: update volt component import

MrYamous:
According to [Livewire V4 upgrade guide](https://livewire.laravel.com/docs/4.x/upgrading#upgrading-volt)

I guess next step will be to do providers change automatically too but I need to dig more to find the best way


calebdw:
Can you add it to the composer set for automatic application for v4 projects?

MrYamous:
> Can you add it to the composer set for automatic application for v4 projects?

I made an update, I hope that was what you mean with "composer set"

PR #469, opened 2026-02-07
Add dependabot.yml

jackbayliss:
This adds a dependabot.yml config which updates github actions 

I've left it pretty basic so you can adjust it as required  - chose weekly to avoid spamming.

see https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/keeping-your-actions-up-to-date-with-dependabot#enabling-dependabot-version-updates-for-actions


jackbayliss:
Tests are failing but not related to this change 🖖🏻

peterfox:
There's no need for composer updates as this is a package library.

jackbayliss:
@peterfox i've updated it, cheers!

PR #477, opened 2026-03-03
Laravel 13 Model Attributes

peterfox:
# Changes

- Adds the Laravel 13 set list
- Adds rules for converting several properties like Hidden, Fillable, Guarded etc.

# Why

These have been confirmed for [Laravel 13](https://laravel-news.com/laravel-13). I wanted to start with these and then move onto the other changes with attributes like console commands and queue jobs.

### DIFF
--- a/src/Rector/Class_/AppendsPropertyToAppendsAttributeRector.php
+++ b/src/Rector/Class_/AppendsPropertyToAppendsAttributeRector.php
@@ -86,6 +86,10 @@ CODE_SAMPLE
 
         $appendsArray = $propertyProperty->default;
 
+        if ($appendsArray->items === []) {
+            return null;
+        }
+
         if (! $this->isArrayOfStrings($appendsArray)) {
             return null;
         }
--- a/src/Rector/Class_/AppendsPropertyToAppendsAttributeRector.php
+++ b/src/Rector/Class_/AppendsPropertyToAppendsAttributeRector.php
@@ -9,7 +9,7 @@ use PhpParser\Node\Arg;
 use PhpParser\Node\Attribute;
 use PhpParser\Node\AttributeGroup;
 use PhpParser\Node\Expr\Array_;
-use PhpParser\Node\Name;
+use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\Scalar\String_;
 use PhpParser\Node\Stmt\Class_;
 use PHPStan\Type\ObjectType;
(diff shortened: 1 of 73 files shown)


src/Rector/Class_/AppendsPropertyToAppendsAttributeRector.php
```diff
@@ -0,0 +1,123 @@
...
+        $appendsArray = $propertyProperty->default;
+
+        if (! $this->isArrayOfStrings($appendsArray)) {
+            return null;
+        }
+
+        // Add attribute to class
+        $node->attrGroups[] = new AttributeGroup([
+            new Attribute(
+                new Name('\Illuminate\Database\Eloquent\Attributes\Appends'),
```
GeniJaho: I think we should be using `new FullyQualified('Illuminate\Database\Eloquent\Attributes\...')` instead of `new Name()` as we do in most of the code of the repo. We do that in ConnectionPropertyToConnectionAttributeRector as well.
We should use FullyQualified in FillablePropertyToFillableAttributeRector, and some other new rules in this PR as well.

tests/Rector/Class_/GuardedPropertyToGuardedAttributeRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace RectorLaravel\Tests\Rector\Class_\GuardedPropertyToGuardedAttributeRector\Fixture;
+
+use Illuminate\Database\Eloquent\Model;
+
+class User extends Model
+{
+    protected $guarded = [
```
GeniJaho: If `$guarded = [];` we would have `#[Guarded([])]`. Is that valid in Laravel 13? Same for $fillable.

src/Rector/Class_/AppendsPropertyToAppendsAttributeRector.php
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        $appendsArray = $propertyProperty->default;
+
+        if (! $this->isArrayOfStrings($appendsArray)) {
+            return null;
+        }
+
+        // Add attribute to class
+        $node->attrGroups[] = new AttributeGroup([
```
GeniJaho: We should add a check whether the target attribute already exists on the class before adding it, in this and other new rules added. I think we can use `PhpAttributeAnalyzer` for this.

GeniJaho (changes requested):

GeniJaho (approved):

masterbater:
v13.1.1
#[Table(table: 'users')]


breaks
Unknown named parameter $table

```php
public function __construct(
    ?string $name = null,
    ?string $key = null,
    ?string $keyType = null,
    ?bool $incrementing = null,
    ?bool $timestamps = null,
    ?string $dateFormat = null
) { }
```

@peterfox

PR #348, opened 2025-06-05
Add RemoveMigrationDocBlockRector rector

glensc:
Cleans up useless phpdoc:

```diff
-/**
- * Run the migrations.
- *
- * @return void
- */
public function up()
{
    // ...
}

-/**
- * Reverse the migrations
- *
- * @return void
- */
public function down()
{
    // ...
}
```

PR #487, opened 2026-03-24
feat: add new RemoveDownMethodFromMigrations as an opt-in rule for removing the down method's of migration

joshdonnell:
As someone who always removes the down() method, I felt this rule could be useful as an opt-in alternative to publishing custom migration stubs via Artisan — with the same end result.

It removes the down() method and its docblock comment from migration files.

Enable it by adding:

``` php
->withRules([
    RemoveDownMethodFromMigrationsRector::class,
]);
````


peterfox:
Thanks @joshdonnell . Is there much value in this over just publishing the stubs and then removing the down method from them compared to having a rule in place?

ghost:
> re much value in this over just publishing the stubs and then removing the down method from them compared to having a rule in place?

Thank you for the reply, I believe there is initially what made me create the PR was my dev team members kept asking for something that would automatically remove the down methods for them as we where already using rector it made more sense than adding to Pint for example. I get that you can modify the stubs but I think having this in rector feels more aligned tooling wise for me personally.

calebdw:
Thanks for putting together a PR, but I'm going to close this as I don't see much value in it (related to https://github.com/driftingly/rector-laravel/pull/348):

- being an optional rule means that most people will not know about it / use it
- there's already rector dead-code rules that remove empty methods (handles the case where `down` is empty)
- if `down` is not empty, then presumably the author meant for it to exist
- if you want to enforce in your own app/codebase, then publishing the stub is the way to go
- outside of packages, migrations should not be kept around for long periods of time and be continually re-ran---the db schema should be squashed and the migrations files should be deleted or archived for future reference. At that point who cares if an extra method exists?

You can simply use this rector in your own application or publish it in your own package if you like.

Thanks for understanding

joshdonnell:
@calebdw 

All makes sense, and as time has passed since this PR went in, I kind of agree. To be honest, in my big projects, I just have the down methods stubbed out anyway. So, really hasn't been too much of an issue. 

Appreciate the time in reviewing 🚀

PR #491, opened 2026-04-01
feat: add rule for Laravel Timestamps property to attribute

yoannblot:
# Context
Laravel 13 introduces the WithoutTimestamps attributes ([source](https://laravel.com/docs/13.x/eloquent#timestamps)).

# Description

This new rule replaces the old property `$timestamps` by the new `#[WithoutTimestamps]` attribute : 

```diff
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Attributes\WithoutTimestamps;

+#[WithoutTimestamps]
 final class User extends Model
 {
-    public $timestamps = false;
 }
```

or, removes the property when the value is true, because that's the default behavior : 
```diff
 use Illuminate\Database\Eloquent\Model;

 final class User extends Model
 {
-    public $timestamps = true;
 }
```

### DIFF
--- a/config/sets/laravel130.php
+++ b/config/sets/laravel130.php
@@ -16,6 +16,7 @@ use RectorLaravel\Rector\Class_\MaxExceptionsPropertyToMaxExceptionsAttributeRec
 use RectorLaravel\Rector\Class_\QueuePropertyToQueueAttributeRector;
 use RectorLaravel\Rector\Class_\TablePropertyToTableAttributeRector;
 use RectorLaravel\Rector\Class_\TimeoutPropertyToTimeoutAttributeRector;
+use RectorLaravel\Rector\Class_\TimestampsPropertyToTimestampsAttributeRector;
 use RectorLaravel\Rector\Class_\TouchesPropertyToTouchesAttributeRector;
 use RectorLaravel\Rector\Class_\TriesPropertyToTriesAttributeRector;
 use RectorLaravel\Rector\Class_\UniqueForPropertyToUniqueForAttributeRector;
@@ -36,6 +37,7 @@ return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->rule(QueuePropertyToQueueAttributeRector::class);
     $rectorConfig->rule(TablePropertyToTableAttributeRector::class);
     $rectorConfig->rule(TimeoutPropertyToTimeoutAttributeRector::class);
+    $rectorConfig->rule(TimestampsPropertyToTimestampsAttributeRector::class);
     $rectorConfig->rule(TouchesPropertyToTouchesAttributeRector::class);
     $rectorConfig->rule(TriesPropertyToTriesAttributeRector::class);
     $rectorConfig->rule(UniqueForPropertyToUniqueForAttributeRector::class);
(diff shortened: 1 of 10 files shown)


src/Rector/Class_/TimestampsPropertyToTimestampsAttributeRector.php
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private function isAnEloquentModelClass(Class_ $class): bool
+    {
+        if ($class->extends === null) {
+            return false;
+        }
+
+        return $this->isName($class->extends, 'Illuminate\Database\Eloquent\Model') || $this->isName($class->extends, 'Model');
+    }
```
peterfox, line 95: this can be replaced by a simple `$this->isObjectType($class, new \PHPStan\Type\ObjectType('Illuminate\Database\Eloquent\Model'))` call.

src/Rector/Class_/TimestampsPropertyToTimestampsAttributeRector.php
```diff
@@ -0,0 +1,127 @@
...
+            return null;
+        }
+
+        $timestampsProperty = $node->getProperty('timestamps');
+        if ($timestampsProperty === null){
+            return null;
+        }
+
+        $this->removePropertyFromClass($node, $timestampsProperty);
+        if($this->isFalseProperty($timestampsProperty)){
```
peterfox, line 81: Equally if you have the property you should be able to do `$this->getType($timestampProperty)->isFalse()->yes()` instead of needing this method implemented.

src/Rector/Class_/TimestampsPropertyToTimestampsAttributeRector.php
```diff
@@ -0,0 +1,127 @@
...
+        if ($this->phpAttributeAnalyzer->hasPhpAttribute($node, 'Illuminate\Database\Eloquent\Attributes\WithoutTimestamps')) {
+            return null;
+        }
+
+        $timestampsProperty = $node->getProperty('timestamps');
+        if ($timestampsProperty === null){
+            return null;
+        }
+
+        $this->removePropertyFromClass($node, $timestampsProperty);
```
peterfox, line 80: This seems wrong and should only happen if the attribute is added because the value is false. It just causes weird edge cases and extends the purepose of the rule beyond it's description.

config/sets/laravel130.php
```diff
@@ -36,6 +37,7 @@
     $rectorConfig->rule(QueuePropertyToQueueAttributeRector::class);
     $rectorConfig->rule(TablePropertyToTableAttributeRector::class);
     $rectorConfig->rule(TimeoutPropertyToTimeoutAttributeRector::class);
+    $rectorConfig->rule(TimestampsPropertyToTimestampsAttributeRector::class);
```
peterfox, line 40: We also changed these to live in their own config instead of laravel130.php

peterfox (changes requested):

yoannblot:
Thanks @peterfox for the review, I will close this PR since I'd like to suggest a better one!

PR #492, opened 2026-04-08
Bump to Rector ^2.4.1 and replace deprecated file to getFile() on rector rule

samsonasik:
Ref 

- https://github.com/rectorphp/rector-src/pull/7962


samsonasik:
Ready to merge for immediate new release :)

samsonasik:
I've re-run rector, should be ok now 👍

Issue #493, opened 2026-04-10
Feature Request: Disable all attribute rules


Muetze42:
Is there a way to disable all attribute rules?

It’s pretty tedious to remove all attribute rules—and future ones—via the config.

Muetze42:
Adds a Laravel 13 set without attribute migrations. Closes #494 .

PR #494, opened 2026-04-11
feat: Add Laravel 13 set without attribute migrations

Muetze42:
Add a Laravel 13 set variant that excludes attribute migration rules.

This splits the current Laravel 13 set into:

- a core set for non-attribute Laravel 13 upgrades
- an attributes-only set for PHP attribute migrations

The default `LARAVEL_130` and `UP_TO_LARAVEL_130` behavior stays unchanged.
New `LARAVEL_130_WITHOUT_ATTRIBUTES` and `UP_TO_LARAVEL_130_WITHOUT_ATTRIBUTES`
constants make it possible to opt out of attribute migrations while still
applying the remaining Laravel 13 upgrade rules.

## Changes

- split `config/sets/laravel130.php` into:
    - `laravel130-core.php`
    - `laravel130-attributes.php`
    - `laravel130-without-attributes.php`
- add `up-to-laravel-130-without-attributes.php`
- add:
    - `LaravelSetList::LARAVEL_130_WITHOUT_ATTRIBUTES`
    - `LaravelLevelSetList::UP_TO_LARAVEL_130_WITHOUT_ATTRIBUTES`
- add tests for the new set variant

## Motivation

Some projects want Laravel 13 upgrade rules, but don?t want Rector to migrate model/job properties to PHP attributes
automatically. This provides a supported alternative without requiring users to maintain a growing skip list.

refs: https://github.com/driftingly/rector-laravel/issues/493

### DIFF
--- a/config/sets/laravel130-attributes.php
+++ b/config/sets/laravel130-attributes.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
+use RectorLaravel\Rector\Class_\AppendsPropertyToAppendsAttributeRector;
+use RectorLaravel\Rector\Class_\BackoffPropertyToBackoffAttributeRector;
+use RectorLaravel\Rector\Class_\ConnectionPropertyToConnectionAttributeRector;
+use RectorLaravel\Rector\Class_\FailOnTimeoutPropertyToFailOnTimeoutAttributeRector;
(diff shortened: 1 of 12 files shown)


peterfox:
This looks good @Muetze42 , I agree, I think it might be best to keep attribute conversion to it's own set as it's not an upgrade as such.

Muetze42:
Ay. The linter.

Now all scripts, `phpstan`, `lint` & `rector-dry-run running` without any errors.

PR #499, opened 2026-04-16
Fix Scope detection false positives

### LINKED ISSUE #498: Attributed scope methods should check that the method is actually a scope
`ScopeNamedClassMethodToScopeAttributedClassMethodRector`, similarly to #379's report on `MakeModelAttributesAndScopesProtectedRector`, does not account for model relations that start with `scope`, i.e. if you have a model that itself is called `Scope`.

I have models named `Scope`, `ScopeType`, and similar, with `HasMany` relations on some models to them, and those are all transformed into attribute based scopes for nonexistent `Type` models and such.

nikspyratos:
Closes #498

TL;DR expands scope checking rule analyzers to check for more markers of a local scope:
1. Method name starts with scope + uppercase char (existing)
2. First parameter is either untyped or typed as `Illuminate\Database\Eloquent\Builder`(new)
3. Return type is absent, void, or Builder (new)

Existing behaviour to check for the method name being only `scope` (#380) is preserved 

In draft until I test this in a real repo soon, unless y'all think it's fine with the test coverage

---

Ah, didn't notice #409 tries to cover the same ground, as there wasn't an issue for it. Mine covers _both_ `ScopeNamedClassMethodToScopeAttributedClassMethodRector` and `MakeModelAttributesAndScopesProtectedRector`, take your pick on which approach is preferred. There are some differences though:

<details> 
<summary>Differences</summary>

- Name pattern check: PR 409 uses `/^scope.+$/` which matches names like `scopedItems`. This PR checks for scope + uppercase char at position 5, which is more precise to Laravel's convention (only `scopeX` where X is uppercase is a valid scope).

- Type validation strategy: PR 409 checks "if return type OR first param type is missing, assume scope". This branch checks each item independently: an untyped first param is accepted, but a non-Builder/non-void return type still disqualifies the method.

- Additional fix: This branch also fixes a pre-existing bug where `str_replace('scope', '', $name)` could remove multiple occurrences of "scope" in a method name (e.g. scopeRescope → re instead of rescope). Replaced with `substr($name, 5)`.

</details>


src/Rector/ClassMethod/MakeModelAttributesAndScopesProtectedRector.php
```diff
@@ -136,12 +136,6 @@ private function isAttributeMethod(ClassMethod $classMethod): bool
...
     {
-        $name = $this->getName($classMethod);
-
-        if ((bool) preg_match('/^scope.+$/', $name)) {
-            return true;
-        }
-
-        return $this->phpAttributeAnalyzer->hasPhpAttribute($classMethod, 'Illuminate\Database\Eloquent\Attributes\Scope');
+        return $this->scopeAnalyzer->isScopeMethod($classMethod);
     }
```
calebdw: there's no need for this method now---just replace with the direct call to the service

src/NodeAnalyzer/ScopeAnalyzer.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function __construct(
+        private NodeNameResolver $nodeNameResolver,
+        private NodeTypeResolver $nodeTypeResolver,
+        private PhpAttributeAnalyzer $phpAttributeAnalyzer,
+    ) {}
+
+    /**
+     * Checks for the "scope" + uppercase char naming convention, a Builder-typed
+     * first parameter, and a void/Builder/untyped return.
```
calebdw: it's uncommon, but scopes can technically return any values---this terminates the fluent query syntax and directly returns the value

src/NodeAnalyzer/ScopeAnalyzer.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function isNamedScope(ClassMethod $classMethod): bool
+    {
+        $name = $this->nodeNameResolver->getName($classMethod);
+
+        if ($name === null || ! str_starts_with($name, 'scope') || strlen($name) <= 5 || ! ctype_upper($name[5])) {
+            return false;
+        }
```
calebdw: this seems more complicated than necessary

```suggestion
        $name = (string) $this->nodeNameResolver->getName($classMethod);

        if ((bool) preg_match('/^scope[A-Z].+$/', $name)) {
            return false;
        }
```

calebdw (changes requested):

src/NodeAnalyzer/ScopeAnalyzer.php
```diff
@@ -0,0 +1,77 @@
...
+
+    private function hasScopeReturnType(ClassMethod $classMethod): bool
+    {
+        if ($classMethod->returnType === null) {
+            return true;
+        }
+
+        if ($classMethod->returnType instanceof Identifier && $classMethod->returnType->toString() === 'void') {
+            return true;
+        }
```
calebdw: look around the codebase and in the rector-src repository---pretty sure there's a better rector way of checking if a method has a void return type

src/NodeAnalyzer/ScopeAnalyzer.php
nikspyratos: Ah, yes true. In that case seems like it'd be simpler to drop the check? Which obviates the need for a better void return check I suppose.

src/NodeAnalyzer/ScopeAnalyzer.php
nikspyratos: Fair enough, regex is simpler, I just got used to not reaching for it most of the time.

nikspyratos:
Thanks for the feedback! Have updated per the suggestions, let me know if that suffices to resolve the comments.

Got around to testing in an affected project and it all worked fine for me, so marking as ready for review properly.

calebdw (approved):

PR #501, opened 2026-05-11
have CarbonSetTestNowToTravelToRector work in pest as well

moshe-autoleadstar:
currently, `CarbonSetTestNowToTravelToRector` will only work on proper test classes and not on Pest tests.
This is because the rector checks `isInClass()`.
This PR relaxes the guard: keep the current behavior when `isInClass()`, and also allow when the scope’s `$this` type is a subtype of `Illuminate\Foundation\Testing\TestCase`


moshe-autoleadstar:
@peterfox can we run the workflows?

moshe-autoleadstar:
@calebdw the failure https://github.com/driftingly/rector-laravel/actions/runs/27674252436/job/82745292191 has nothing to do with this PR..

src/Rector/StaticCall/CarbonSetTestNowToTravelToRector.php
```diff
@@ -107,6 +104,21 @@ public function refactor(Node $node): ?MethodCall
         );
     }
 
+    private function isInLaravelTestCaseScope(Scope $scope): bool
+    {
+        $testCaseClass = 'Illuminate\Foundation\Testing\TestCase';
```
calebdw: we typically move this to a private const on the class

src/Rector/StaticCall/CarbonSetTestNowToTravelToRector.php
```diff
@@ -107,6 +104,21 @@ public function refactor(Node $node): ?MethodCall
...
+
+        if ($scope->isInClass()) {
+            return $scope->getClassReflection()->isSubclassOfClass($this->reflectionProvider->getClass($testCaseClass));
```
calebdw: nit: this is a little more concise

```suggestion
            return $scope->getClassReflection()->is($testCaseClass);
```

src/Rector/StaticCall/CarbonSetTestNowToTravelToRector.php
```diff
@@ -107,6 +104,21 @@ public function refactor(Node $node): ?MethodCall
...
+        }
+
+        // Pest / other closures: `@param-closure-this` on the test runner makes `$this` a TestCase.
+        $thisType = $scope->getType(new Variable('this'));
```
calebdw, line 113: does this return a null that needs to be handled?

src/Rector/StaticCall/CarbonSetTestNowToTravelToRector.php
```diff
@@ -107,6 +104,21 @@ public function refactor(Node $node): ?MethodCall
...
 
+    private function isInLaravelTestCaseScope(Scope $scope): bool
+    {
+        $testCaseClass = 'Illuminate\Foundation\Testing\TestCase';
+
+        if ($scope->isInClass()) {
+            return $scope->getClassReflection()->isSubclassOfClass($this->reflectionProvider->getClass($testCaseClass));
+        }
+
+        // Pest / other closures: `@param-closure-this` on the test runner makes `$this` a TestCase.
```
calebdw, line 112: can we add a test for this?

src/Rector/StaticCall/CarbonSetTestNowToTravelToRector.php
moshe-autoleadstar: No. `Scope::getType()` always returns a `Type`, never `null`. The closure branch keeps `ObjectType::isSuperTypeOf()` which safely returns `false` for unrelated types.

Also, if I do try to handle the null case, I get a phpstan error:
```
./vendor/bin/phpstan analyse src/Rector/StaticCall/CarbonSetTestNowToTravelToRector.php
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ----------------------------------------------------------------------------------------------- 
  Line   CarbonSetTestNowToTravelToRector.php                                                           
 ------ ----------------------------------------------------------------------------------------------- 
  114    Strict comparison using === between PHPStan\Type\Type and null will always evaluate to false.  
         🪪  identical.alwaysFalse                                                                      
 ------ ----------------------------------------------------------------------------------------------- 
                                                                                                                        
 [ERROR] Found 1 error                                                                                                  
```

So, the way we have it is correct.

calebdw (approved):

calebdw:
Unfortunately, it does not look like I can merge when ci is failing

moshe-autoleadstar:
@calebdw ok. ive made the changes that are unrelated to this PR but making the CI fail. now everything passes
image

PR #504, opened 2026-05-15
Bump structarmed to ^0.5

samsonasik:


samsonasik:
All green, latest 0.5.x bring various performance improvement with parallel analysis. Let's merge ;)

PR #506, opened 2026-05-19
Feat: missing eloquent attribute rectors

MrPunyapal:
## Rules Added
- `VisiblePropertyToVisibleAttributeRector`
- `DateFormatPropertyToDateFormatAttributeRector`
- `WithoutTimestampsPropertyToWithoutTimestampsAttributeRector`
- `WithoutIncrementingPropertyToWithoutIncrementingAttributeRector`
- `CollectedByPropertyToCollectedByAttributeRector`

### VisiblePropertyToVisibleAttributeRector
Before:
```php

class User extends Model
{
    protected $visible = ['id', 'name'];
}
```
After:
```php
use Illuminate\Database\Eloquent\Attributes\Visible;

#[Visible(['id', 'name'])]
class User extends Model
{
}
```

### DateFormatPropertyToDateFormatAttributeRector
Before:
```php

class User extends Model
{
    protected $dateFormat = 'U';
}
```
After:
```php
use Illuminate\Database\Eloquent\Attributes\DateFormat;

#[DateFormat('U')]
class User extends Model
{
}
```

### WithoutTimestampsPropertyToWithoutTimestampsAttributeRector
Before:
```php

class User extends Model
{
    public $timestamps = false;
}
```
After:
```php
use Illuminate\Database\Eloquent\Attributes\WithoutTimestamps;

#[WithoutTimestamps]
class User extends Model
{
}
```

### WithoutIncrementingPropertyToWithoutIncrementingAttributeRector
Before:
```php

class User extends Model
{
    public $incrementing = false;
}
```
After:
```php
use Illuminate\Database\Eloquent\Attributes\WithoutIncrementing;

#[WithoutIncrementing]
class User extends Model
{
}
```

### CollectedByPropertyToCollectedByAttributeRector
Before:
```php
use App\Models\Collections\UserCollection;
use Illuminate\Database\Eloquent\Collection;

class User extends Model
{
    /**
     * @param array<int, \Illuminate\Database\Eloquent\Model> $models
     * @return \Illuminate\Database\Eloquent\Collection<int, \Illuminate\Database\Eloquent\Model>
     */
    public function newCollection(array $models = []): Collection
    {
        return new UserCollection($models);
    }
}
```
After:
```php
use App\Models\Collections\UserCollection;
use Illuminate\Database\Eloquent\Attributes\CollectedBy;

#[CollectedBy(UserCollection::class)]
class User extends Model
{
}
```

### DIFF
--- a/src/Rector/Class_/VisiblePropertyToVisibleAttributeRector.php
+++ b/src/Rector/Class_/VisiblePropertyToVisibleAttributeRector.php
@@ -1,0 +1,134 @@
+<?php
+
+declare(strict_types=1);
+
+namespace RectorLaravel\Rector\Class_;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Attribute;
+use PhpParser\Node\AttributeGroup;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\Scalar\String_;
+use PhpParser\Node\Stmt\Class_;
+use PHPStan\Type\ObjectType;
+use Rector\Php80\NodeAnalyzer\PhpAttributeAnalyzer;
+use RectorLaravel\AbstractRector;
+use RectorLaravel\Tests\Rector\Class_\VisiblePropertyToVisibleAttributeRector\VisiblePropertyToVisibleAttributeRectorTest;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see VisiblePropertyToVisibleAttributeRectorTest
+ */
+final class VisiblePropertyToVisibleAttributeRector extends AbstractRector
+{
+    public function __construct(private readonly PhpAttributeAnalyzer $phpAttributeAnalyzer) {}
+
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition(
+            'Changes model visible property to use the Visible attribute',
+            [new CodeSample(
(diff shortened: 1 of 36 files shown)


src/Rector/Class_/VisiblePropertyToVisibleAttributeRector.php
```diff
@@ -0,0 +1,134 @@
...
+                return false;
+            }
+
+            if (! $item->value instanceof String_) {
+                return false;
+            }
+        }
+
+        return true;
+    }
```
peterfox, line 132: Might be better to do the check `$this->getType($item->value)->isString()->yes()` instead of checking the node as there might be constants used or something silly like that. It's unlikely but doesn't hurt to use PHPStan's analysis instead of checking the node types.

peterfox:
This looks good at a quick glance. Thank you @MrPunyapal.

Something to consider, and I think I need to go back and look at how this has been done with the rules I made + how Laravel itself handles it. But I'm not sure what these rules should do if both a attribute exists and the method/property, but the method/properties's resulting attributes would be difficult to the current attribute.

e.g. 

```php
use App\Models\Collections\UserCollection;
use Illuminate\Database\Eloquent\Collection;

#[CollectedBy(SomeOtherCollection::class)]
class User extends Model
{
    /**
     * @param array<int, \Illuminate\Database\Eloquent\Model> $models
     * @return \Illuminate\Database\Eloquent\Collection<int, \Illuminate\Database\Eloquent\Model>
     */
    public function newCollection(array $models = []): Collection
    {
        return new UserCollection($models);
    }
}
```

Should it:

a. override the old attribute and remove the method
b. just ignore it

MrPunyapal:
@peterfox can you check now 👀

peterfox (approved):

PR #507, opened 2026-05-19
Feat: missing queue attribute rectors

MrPunyapal:
## Rules Added
- DelayPropertyToDelayAttributeRector
- DeleteWhenMissingModelsPropertyToDeleteWhenMissingModelsAttributeRector

### DelayPropertyToDelayAttributeRector
Before:
```php
use Illuminate\Contracts\Queue\ShouldQueue;

final class SendEmailJob implements ShouldQueue
{
    public $delay = 10;
}
```
After:
```php
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Queue\Attributes\Delay;

#[Delay(10)]
final class SendEmailJob implements ShouldQueue
{
}
```

### DeleteWhenMissingModelsPropertyToDeleteWhenMissingModelsAttributeRector
Before:
```php
use Illuminate\Contracts\Queue\ShouldQueue;

final class SendEmailJob implements ShouldQueue
{
    public $deleteWhenMissingModels = true;
}
```
After:
```php
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Queue\Attributes\DeleteWhenMissingModels;

#[DeleteWhenMissingModels]
final class SendEmailJob implements ShouldQueue
{
}
```


tests/Rector/Class_/DeleteWhenMissingModelsPropertyToDeleteWhenMissingModelsAttributeRector/Fixture/skip_not_final.php.inc
peterfox, line 1: I think this will be okay as personally I prefer that we only change classes that are marked as final but I'm not sure if this is consistent against all the attribute rules.

src/Rector/Class_/DeleteWhenMissingModelsPropertyToDeleteWhenMissingModelsAttributeRector.php
```diff
@@ -0,0 +1,112 @@
...
+
+        $propertyProperty = $property->props[0];
+        if ($propertyProperty->default === null) {
+            return null;
+        }
+
+        $value = $propertyProperty->default;
+        if (! $value instanceof ConstFetch || $value->name->toLowerString() !== 'true') {
+            return null;
+        }
```
peterfox: we probably need some form of check to make sure the value is a boolean true value.

```php
$this->getType($value)->isTrue()->yes();
```

Otherwise the attribute will fail.

src/Rector/Class_/DelayPropertyToDelayAttributeRector.php
```diff
@@ -0,0 +1,112 @@
...
+        if (! $property->isPublic()) {
+            return null;
+        }
+
+        $propertyProperty = $property->props[0];
+        if ($propertyProperty->default === null) {
+            return null;
+        }
+
+        $value = $propertyProperty->default;
```
peterfox, line 94: we probably need some form of check to make sure the value is an integer value.

```php
$this->getType($value)->isInt()->yes();
```

Otherwise the attribute will fail.

tests/Rector/Class_/DeleteWhenMissingModelsPropertyToDeleteWhenMissingModelsAttributeRector/Fixture/skip_not_final.php.inc
MrPunyapal: I think it is consistent?

- class: [`RectorLaravel\Rector\Class_\FailOnTimeoutPropertyToFailOnTimeoutAttributeRector`](../src/Rector/Class_/FailOnTimeoutPropertyToFailOnTimeoutAttributeRector.php)

```diff
 use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Queue\Attributes\FailOnTimeout;

+#[FailOnTimeout(true)]
 final class ProcessPodcast implements ShouldQueue
 {
-    public $failOnTimeout = true;
 }
```

Image

PR #512, opened 2026-05-31
Update boundwize/structarmed version to ^0.9

samsonasik:


samsonasik:
PHPStan notice is unrelated, I will create separate PR for it.

PR #515, opened 2026-05-31
Fix: guard single-arg where() against array arguments in WhereNullComparisonToWhereNullRector

### LINKED ISSUE #514: WhereNullComparisonToWhereNullRector corrupts where() calls that take an array argument
> **Note:** The following was written by an LLM, but verified to be correct by a human.

### Description

`WhereNullComparisonToWhereNullRector`'s single-argument branch rewrites `where($x)` → `whereNull($x)` unconditionally, with no type guard on the argument. It assumes a one-arg `where()` is always `where('column')`. But the array form `where(['col' => $value, ...])` is _also_ a single-argument call, so it gets converted to `whereNull(['col' => $value, ...])`, which changes the query's semantics: `where([...])` is an AND-of-equality filter, whereas `whereNull()` expects column name(s) and checks for NULL.

### Offending code

`src/Rector/MethodCall/WhereNullComparisonToWhereNullRector.php`, lines 62–68:

```php
if (count($args) === 1) {
    return $this->nodeFactory->createMethodCall(
        $node->var,
        'whereNull',
        [$args[0]]
    );
}
```

The 2-arg and 3-arg branches correctly guard with `if ($type->isNull()->no()) { return null; }` (lines 85–91), but the single-arg branch skips that check entirely.

### Reproduction

Before:

```php
User::query()->where(['name' => 'foo', 'active' => true])->exists();
```

Rector output (WRONG):

```php
User::query()->whereNull(['name' => 'foo', 'active' => true])->exists();
```

Expected: unchanged — an array-form `where([...])` is not a null comparison.

### Suggested fix

In `refactor()`, make the `count($args) === 1` branch bail unless the argument is string-typed, mirroring the null guard already used by the other branches, e.g.:

```php
if (count($args) === 1) {
    if (! $args[0] instanceof Arg || ! $this->getType($args[0]->value)->isString()->yes()) {
        return null;
    }
    return $this->nodeFactory->createMethodCall($node->var, 'whereNull', [$args[0]]);
}
```

A regression fixture with an array argument (expected: no change) would also catch this — the current fixtures only cover string-column args.

### Environment

* driftingly/rector-laravel: 2.4.0 (latest)
* rector/rector: 2.4.5
* PHP: 8.5
* Laravel: 13

Plytas:
> **Note:** The following was written by an LLM, but verified to be correct by a human.

Fixes #514.

## Summary

- The single-arg branch of `WhereNullComparisonToWhereNullRector::refactor()` was unconditionally rewriting `where($x)` → `whereNull($x)` with no type guard. This corrupts array-form `where(['col' => $value, ...])` calls, changing an AND-of-equality filter into a semantically different `whereNull()` call.
- The fix adds a string-type guard (`isString()->yes()`) matching the idiom already used by the 2-arg and 3-arg branches.
- A new skip fixture covers the regressing case; all 622 tests pass.

## Changes

- `src/Rector/MethodCall/WhereNullComparisonToWhereNullRector.php` — add early-return guard in the `count($args) === 1` branch
- `tests/Rector/MethodCall/WhereNullComparisonToWhereNullRector/Fixture/skip_array_argument.php.inc` — new skip fixture for array-argument form

## Test plan

- [ ] New fixture `skip_array_argument.php.inc` fails before the fix, passes after
- [ ] Existing positive fixtures (`where('foobar')`, `where('foo', null)`, etc.) still convert correctly
- [ ] Full suite: `vendor/bin/phpunit` — 622 tests, 807 assertions, all green

### DIFF
--- a/src/Rector/MethodCall/WhereNullComparisonToWhereNullRector.php
+++ b/src/Rector/MethodCall/WhereNullComparisonToWhereNullRector.php
@@ -60,6 +60,10 @@ CODE_SAMPLE
         $type = null;
 
         if (count($args) === 1) {
+            if (! $args[0] instanceof Arg || ! $this->getType($args[0]->value)->isString()->yes()) {
+                return null;
+            }
+
             return $this->nodeFactory->createMethodCall(
                 $node->var,
                 'whereNull',
--- a/tests/Rector/MethodCall/WhereNullComparisonToWhereNullRector/Fixture/skip_array_argument.php.inc
+++ b/tests/Rector/MethodCall/WhereNullComparisonToWhereNullRector/Fixture/skip_array_argument.php.inc
@@ -1,0 +1,14 @@
+<?php
+
+namespace RectorLaravel\Tests\Rector\MethodCall\AssertStatusToAssertMethodRector\Fixture;
+
+use Illuminate\Contracts\Database\Query\Builder;
+
+class SkipArrayArgument
+{
+    public function test(Builder $query)
+    {
+        $query->where(['name' => 'foo', 'active' => true])->exists();
+    }
+}
+?>


calebdw:
Thanks!

PR #516, opened 2026-06-02
Skip superglobal assignments in Request, Env and Server rectors

shaffe-fr:
The `RequestVariablesToRequestFacadeRector`, `EnvVariableToEnvHelperRector` and `ServerVariableToRequestFacadeRector` rules currently transform superglobal assignments, which produces invalid code.

**Before this fix:**

`php
// This valid code:
$_POST = $data;
$_ENV['APP_NAME'] = 'MyApp';

// Was wrongly transformed into:
\Illuminate\Support\Facades\Request::post() = $data;         // fatal error
\Illuminate\Support\Env::get('APP_NAME') = 'MyApp';           // fatal error
`

**After this fix**, assignments to superglobals are left untouched while reads are still correctly transformed:

`php
// Assignments: left alone
$_POST = $data;
$_POST['key'] = $value;
$_ENV['APP_NAME'] = 'MyApp';

// Reads: still transformed
$var = $_POST['key'];    // => Request::post('key')
$env = $_ENV['APP_DEBUG']; // => Env::get('APP_DEBUG')
`

Also renames `IS_IN_SERVER_VARIABLE` to `IS_IN_SUPERGLOBAL_ASSIGN` since the visitor now covers all superglobals.

### DIFF
--- a/src/NodeVisitor/ArrayDimFetchContextNodeVisitor.php
+++ b/src/NodeVisitor/ArrayDimFetchContextNodeVisitor.php
@@ -21,8 +21,10 @@ use Rector\PhpDocParser\NodeTraverser\SimpleCallableNodeTraverser;
 final class ArrayDimFetchContextNodeVisitor extends NodeVisitorAbstract implements DecoratingNodeVisitorInterface
 {
     public const string IS_INSIDE_ARRAY_DIM_FETCH_WITH_DIM_NOT_SCALAR = 'is_inside_array_dim_fetch_with_dim_not_scalar';
-    public const string IS_IN_SERVER_VARIABLE = 'is_in_server_variable';
+    public const string IS_IN_SUPERGLOBAL_ASSIGN = 'is_in_superglobal_assign';
     public const string IS_INSIDE_ARRAY_DIM_FETCH_WITH_DIM_NOT_EXPR = 'is_inside_array_dim_fetch_with_dim_not_expr';
+
+    private const array SUPERGLOBAL_NAMES = ['_SERVER', '_GET', '_POST', '_REQUEST', '_ENV'];
 
     public function __construct(
         private readonly NodeNameResolver $nodeNameResolver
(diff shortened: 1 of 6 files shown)


src/NodeVisitor/ArrayDimFetchContextNodeVisitor.php
```diff
@@ -75,4 +73,23 @@ public function enterNode(Node $node)
...
         return null;
     }
+
+    private function isSuperglobalAssign(Node $node): bool
+    {
+        if (! $node instanceof Assign) {
+            return false;
+        }
+
+        $names = ['_SERVER', '_GET', '_POST', '_REQUEST', '_ENV'];
```
calebdw: can you please make this a private constant array?

src/Rector/ArrayDimFetch/EnvVariableToEnvHelperRector.php
```diff
@@ -41,6 +43,14 @@ public function getNodeTypes(): array
      */
     public function refactor(Node $node): ?StaticCall
     {
+        if ($node->getAttribute(AttributeKey::IS_BEING_ASSIGNED) === true) {
+            return null;
+        }
+
+        if ($node->getAttribute(ArrayDimFetchContextNodeVisitor::IS_IN_SUPERGLOBAL_ASSIGN) === true) {
+            return null;
+        }
```
calebdw: why do we need both of these? Is the first condition not enough?

shaffe-fr:
Thanks for the review! I fixed both comments in the latest push.

Regarding ServerVariableToRequestFacadeRector: the assignment skip is already covered by the existing `skip_assign.php.inc` fixture.

src/Rector/ArrayDimFetch/RequestVariablesToRequestFacadeRector.php
```diff
@@ -67,6 +68,14 @@ public function getNodeTypes(): array
      */
     public function refactor(Node $node): StaticCall|NotIdentical|null
     {
+        if ($node->getAttribute(AttributeKey::IS_BEING_ASSIGNED) === true) {
+            return null;
+        }
+
+        if ($node->getAttribute(ArrayDimFetchContextNodeVisitor::IS_IN_SUPERGLOBAL_ASSIGN) === true) {
+            return null;
+        }
```
calebdw, line 80: same thing goes here---is the `IS_BEING_ASSIGNED` not enough to detect and skip this? I'm confused why we really need the `ArrayDimFetchContextNodeVisitor::IS_IN_SUPERGLOBAL_ASSIGN`?

src/Rector/ArrayDimFetch/ServerVariableToRequestFacadeRector.php
```diff
@@ -51,7 +51,7 @@ public function refactor(Node $node): ?StaticCall
             return null;
         }
 
-        if ($node->getAttribute(ArrayDimFetchContextNodeVisitor::IS_IN_SERVER_VARIABLE) === true) {
+        if ($node->getAttribute(ArrayDimFetchContextNodeVisitor::IS_IN_SUPERGLOBAL_ASSIGN) === true) {
```
calebdw, line 54: can this just be?

```suggestion
        if ($node->getAttribute(AttributeKey::IS_BEING_ASSIGNED) === true) {
```

src/Rector/ArrayDimFetch/RequestVariablesToRequestFacadeRector.php
shaffe-fr: Good question. `IS_BEING_ASSIGNED` only covers the direct assignment target (the `ArrayDimFetch` in `['key'] = `). But this rule also listens on `Variable` nodes, and the inner `` variable doesn't get `IS_BEING_ASSIGNED`—only the outer ArrayDimFetch does. Without the second check, `['key'] = ` would transform the inner `` into `Request::post()`. Same thing for isset/unset/interpolated strings which don't set `IS_BEING_ASSIGNED` at all.

calebdw:
Thanks!

PR #519, opened 2026-06-03
fix: LARAVEL_STATIC_TO_INJECTION to point to correct translation methods.

Andyuu:
The trans and transChoice methods do not exist on the Illuminate\Translation\Translator class. The correct methods are get() and choice().


calebdw:
Thanks!

PR #525, opened 2026-06-27
Feat: Add Factory `has()`/`for()` to Magic Method Rector

MizouziE:
See (now removed, as per repo convention) description below:

```php
/**
 * Rewrites factory ->has()/->for() calls into their magic method equivalents, folding any
 * count and literal state into the arguments (e.g. ->has(Variation::factory()->times(3)) becomes
 * ->hasVariations(3)). Genuine relationships with simple factories qualify, including polymorphic
 * has-side relations; only the inverse morphTo is excluded.
 *
 * @see FactoryHasForToMagicMethodRectorTest
 */
```

Hi, I work with an ancient (18 years old) PHP codebase which has been converted to use Laravel over the last 5+ years so a lot of it has very old ways of working still, but we are trying our hardest to be progressive and make use of the multitude of Laravel 13 conveniences. We had previously decided that, going forward, we would make more use of the eloquent magic methods for factory relationship definitions as we often find complex nested factories to be quite distracting when reading the setup for tests.

However, we all attended Laravel Live London recently and it was brought to our attention that we could use Rector (better than we already had been already) to help enforce it going forward **and** convert existing cases in one fell swoop.

This rule in the PR comes from the custom rule I got working with our 4000+ PHPUnit tests (+ a lot of Dusk tests) and we were all quite happy with the result, so I have tried to share it here and make it fit better with Rector-Laravel conventions.

Please do let me know if I can make further changes or take different approaches anywhere.

### Possible concerns

- There are an awful lot of fixtures, but it's due to there being an awful lot of edge cases that I found while developing this. I believe that I came to a sensible and acceptable position before finalising
- I had to add some missing relationship stubs: I would be happy to add those in a separate PR if it's preferred?
- I used British English for one of the self contained helper method names (cos I'm Brit-ish, innit)
- I regenerated the docs as a final step as my understanding of the CI run is that it would not accept this without it (?): If I got that wrong I left it as the final commit so it could be easily reverted

Thanks for your time and a great package. It's value is immeasurable to teams like mine!

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1,4 +1,4 @@
-# 122 Rules Overview
+# 125 Rules Overview
 
 ## AbortIfRector
 
@@ -69,7 +69,7 @@ Add generic Builder return type to scopes in child of `Illuminate\Database\Eloqu
  use App\Post;
  use Illuminate\Database\Eloquent\Model;
  use Illuminate\Database\Eloquent\Builder;
- 
+
  class Post extends Model
  {
 +    /**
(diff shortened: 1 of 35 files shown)


MizouziE:
Closing as I'd like to refine and rework this a bit

PR #526, opened 2026-06-29
Add Eloquent relation stubs for HasOne, MorphMany, MorphOneOrMany, MorphToMany

MizouziE:
Adds test stubs for four Eloquent relation classes that aren't currently covered:
- HasOne
- MorphMany
- MorphOneOrMany
- MorphToMany

They mirror the existing stubs under `stubs/Illuminate/Database/Eloquent/Relations/` and let fixtures declare these relation return types so PHPStan can resolve them in rule tests.

Split into its own PR so it can land independently of [FactoryHasForToMagicMethod](https://github.com/driftingly/rector-laravel/pull/525), the rule that depends on these stubs to help keep that rule's diff focused on the rule itself.

### DIFF
--- a/stubs/Illuminate/Database/Eloquent/Relations/HasOne.php
+++ b/stubs/Illuminate/Database/Eloquent/Relations/HasOne.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace Illuminate\Database\Eloquent\Relations;
+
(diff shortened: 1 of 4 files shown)


calebdw:
Thanks, but in the future I would like to see changes like this on the PR where they are actually used :+1:

PR #528, opened 2026-06-30
Add Laravel 13 ObservedBy observer migration rectors

ivanvasechko:
## Summary

Adds Laravel 13 support for migrating manual model observer registration such as:

```php
User::observe(UserObserver::class);
```

into model attributes:

```php
#[ObservedBy([UserObserver::class])]
```

## What changed

- adds `ObserveCallsToObservedByAttributeRector` to add or merge `#[ObservedBy([...])]` on Eloquent models
- adds `RemoveModelObserveCallsFromBootRector` to remove safe direct top-level `observe()` calls from `boot()` and `booted()`
- supports `User::observe(...)`, `self::observe(...)`, and `static::observe(...)`
- removes empty `boot()` / `booted()` methods after cleanup, including cases with method attributes such as `#[Override]`
- preserves unrelated logic inside boot methods
- skips nested registrations and unsupported existing `ObservedBy(...)` shapes conservatively
- wires both rules into the Laravel 13 attributes set
- documents both rules in `docs/rector_rules_overview.md`

## Precision / safety rules

This change is intentionally conservative:

- only direct top-level `observe()` statements are removed
- nested registrations inside conditionals, loops, or closures are left untouched
- cleanup only runs when the target model can be updated safely
- existing `ObservedBy(...)` attributes are merged and deduplicated when they are in a supported form
- unsupported existing attribute shapes are left alone

## Implementation notes

- project-wide observer discovery now uses Rector-native file discovery instead of custom recursive scanning
- fixture companion files were split into PSR-4-compliant files so repository tooling keeps working out of the box
- added minimal stubs needed for analysis/tests around `ObservedBy` and Eloquent model observation

Tiny confession: rule started a bit vibe-coded, but ended up heavily constrained and aggressively tested. Cave paintings, but with guard rails.

## Tests

Added targeted PHPUnit coverage for:

- simple observe-to-attribute conversion
- merging with existing `ObservedBy`
- skipping duplicate existing attributes
- skipping nested observe registrations
- skipping non-models
- removing direct observe calls while preserving other boot logic
- removing empty `boot()` methods entirely
- removing empty `booted()` methods entirely
- handling `static::observe(...)` inside model `booted()`
- keeping unsupported existing `ObservedBy(...)` forms untouched
- Laravel 13 set-level coverage

Also ran the full local CI-equivalent verification on the repo's original tool configuration:

- `composer validate`
- `composer lint`
- `composer rector-dry-run`
- `composer phpstan`
- `composer structarmed`
- `vendor/bin/phpunit`


calebdw:
Things may have been updated since the last time I dealt with this, but I have run into situations where converting to the `ObservedBy` attribute has broken code (mainly when dealing with model inheritance). Please double check to see if this is still the case, and if so, then please write a test to ensure such a case is skipped:


```php
class File extends Model
{
    #[Override]
    protected static function booted(): void
    {
        // cant' use ObserveredBy attribute here because
        // it doesn't work with inherited classes
        static::observe(FileObserver::class);
    }
}

class UserAvatar extends File {} // please ensure that the hooks defined in FileObserver are actually called when dealing with UserAvatar instances
```

calebdw:
>  removes empty boot() / booted() methods after cleanup, including cases with method attributes such as #[Override]

This rule should not be concerned about this cleanup---there's other rector dead-code rules that perform this and we should let them clean it up

src/Rector/ClassMethod/RemoveModelObserveCallsFromBootRector.php
```diff
@@ -0,0 +1,116 @@
...
+use RectorLaravel\NodeAnalyzer\ObservedByAnalyzer;
+use RectorLaravel\Tests\Rector\ClassMethod\RemoveModelObserveCallsFromBootRector\RemoveModelObserveCallsFromBootRectorTest;
+use RectorLaravel\ValueObject\ObservedByRegistration;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see RemoveModelObserveCallsFromBootRectorTest
+ */
+final class RemoveModelObserveCallsFromBootRector extends AbstractRector
```
calebdw, line 22: Having two separate, but dependent rules like this is a bad architecture---we should only have one rector rule that removes the call from boot/booted and adds the ObservedBy attribute, this will have to operate at the class level

What if someone skipped the other rule but forgot to skip this? Now their codebase is broken

ivanvasechko:
Feedback makes sense. I might get back to it later, but I will close this out for now.

PR #529, opened 2026-07-07
Add AddGenericBuilderToScopesRector 

MrPunyapal:
This rule will simply add generic phpdocs to Laravel scopes.

```diff
class User extends Model
{
    /**
+    * @param Builder<User> $query
+    * @return Builder<User>
     */
    public function scopePopular(Builder $query): Builder
    {
        return $query->where('votes', '>', 100);
    }
}
```

### DIFF
--- a/src/Rector/ClassMethod/AddGenericBuilderToScopesRector.php
+++ b/src/Rector/ClassMethod/AddGenericBuilderToScopesRector.php
@@ -11,6 +11,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\PhpDocParser\Ast\PhpDoc\ParamTagValueNode;
 use PHPStan\PhpDocParser\Ast\PhpDoc\ReturnTagValueNode;
 use PHPStan\PhpDocParser\Ast\Type\GenericTypeNode;
+use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\TypeNode;
 use PHPStan\Reflection\ClassReflection;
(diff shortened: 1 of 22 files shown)


calebdw:
The generics should be `Builder<self>` if the class is final, or `Builder<static>` if not

MrPunyapal:
@calebdw, I find in most places it is the model name instead of self or static 
 What do you say, @peterfox?

MrPunyapal:
Anyway, I've done that, but if it is not needed, we can revert the last commit. Also, CI is unrelated. I can do a separate PR if needed.

src/Rector/ClassMethod/AddGenericBuilderToScopesRector.php
```diff
@@ -0,0 +1,293 @@
...
+
+        if (count($node->params) < 1) {
+            return null;
+        }
+
+        $methodName = $this->getName($node);
+        $hasScopeAttribute = $this->phpAttributeAnalyzer->hasPhpAttribute(
+            $node,
+            'Illuminate\Database\Eloquent\Attributes\Scope',
+        );
```
calebdw: a lot of this can be cleaned up by using the `ScopeAnalyzer`/`ModelAnalyzer`

MrPunyapal: used it

calebdw:
Looks good, unfortunately I can't merge without ci passing (if you don't mind, please just add the fix in a separate commit)

calebdw:
Thanks!

PR #533, opened 2026-07-19
Bump to StructArmed ^0.15 and enable MustBeFinalRule

samsonasik:
StructArmed 0.15.0 introduce `ExtendedClassAwareRuleInterface` that makes `MustBeFinalRule` safely be used to skip if there is class that extend target class.

This PR bump StructArmed to `^0.15` and apply the `MustBeFinalRule` with single command:

```bash
vendor/bin/structarmed analyze --fix
```


samsonasik:
@peterfox ready to merge 👍

PR #537, opened 2026-07-24
Add RouteKeyMethodToRouteKeyAttributeRector 

MrPunyapal:
This PR adds a rule to apply the new RouteKey rule.

Example.
```diff
 use Illuminate\Database\Eloquent\Model;

+#[\Illuminate\Database\Eloquent\Attributes\RouteKey('slug')]
 class Post extends Model
 {
-    public function getRouteKeyName(): string
-    {
-        return 'slug';
-    }
 }
```

### DIFF
--- a/src/Rector/Class_/RouteKeyMethodToRouteKeyAttributeRector.php
+++ b/src/Rector/Class_/RouteKeyMethodToRouteKeyAttributeRector.php
@@ -1,0 +1,116 @@
+<?php
+
+declare(strict_types=1);
+
+namespace RectorLaravel\Rector\Class_;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Attribute;
+use PhpParser\Node\AttributeGroup;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\Scalar\String_;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Return_;
+use PHPStan\Type\ObjectType;
+use Rector\Php80\NodeAnalyzer\PhpAttributeAnalyzer;
+use RectorLaravel\AbstractRector;
+use RectorLaravel\Tests\Rector\Class_\RouteKeyMethodToRouteKeyAttributeRector\RouteKeyMethodToRouteKeyAttributeRectorTest;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see RouteKeyMethodToRouteKeyAttributeRectorTest
+ */
+final class RouteKeyMethodToRouteKeyAttributeRector extends AbstractRector
+{
+    public function __construct(private readonly PhpAttributeAnalyzer $phpAttributeAnalyzer) {}
+
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition(
+            'Changes model getRouteKeyName() method to use the RouteKey attribute',
+            [new CodeSample(
+                <<<'CODE_SAMPLE'
+use Illuminate\Database\Eloquent\Model;
+
+class Post extends Model
+{
+    public function getRouteKeyName(): string
+    {
+        return 'slug';
+    }
+}
+CODE_SAMPLE,
+                <<<'CODE_SAMPLE'
+use Illuminate\Database\Eloquent\Model;
+
+#[\Illuminate\Database\Eloquent\Attributes\RouteKey('slug')]
+class Post extends Model
+{
+}
+CODE_SAMPLE
+            )]
+        );
+    }
+
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [Class_::class];
+    }
+
+    /**
+     * @param  Class_  $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->isObjectType($node, new ObjectType('Illuminate\Database\Eloquent\Model'))) {
+            return null;
+        }
+
+        if ($this->phpAttributeAnalyzer->hasPhpAttribute($node, 'Illuminate\Database\Eloquent\Attributes\RouteKey')) {
+            return null;
+        }
+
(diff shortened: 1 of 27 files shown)


src/Rector/Class_/RouteKeyMethodToRouteKeyAttributeRector.php
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        $returnExpr = $this->resolveReturnString($routeKeyMethod);
+        if (! $returnExpr instanceof String_) {
+            return null;
+        }
+
+        $node->attrGroups[] = new AttributeGroup([
+            new Attribute(new FullyQualified('Illuminate\Database\Eloquent\Attributes\RouteKey'), [new Arg($returnExpr)]),
+        ]);
```
calebdw: This entire `resolveReturnSring` method can be replaced with rectors `ReturnNodeFinder` service

```suggestion
        $return = $this->returnNoteFinder->findOnlyReturnWithExpr($routeKeyMethod);
        if (! $return?->expr instanceof String_) {
            return null;
        }

        $node->attrGroups[] = new AttributeGroup([
            new Attribute(new FullyQualified('Illuminate\Database\Eloquent\Attributes\RouteKey'), [new Arg($return->expr)]),
        ]);
```

src/Rector/Class_/RouteKeyMethodToRouteKeyAttributeRector.php
```diff
@@ -0,0 +1,116 @@
...
+use Rector\Php80\NodeAnalyzer\PhpAttributeAnalyzer;
+use RectorLaravel\AbstractRector;
+use RectorLaravel\Tests\Rector\Class_\RouteKeyMethodToRouteKeyAttributeRector\RouteKeyMethodToRouteKeyAttributeRectorTest;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see RouteKeyMethodToRouteKeyAttributeRectorTest
+ */
+final class RouteKeyMethodToRouteKeyAttributeRector extends AbstractRector
```
calebdw: Please implement the new rector `ComposerPackageConstraintInterface`: https://github.com/rectorphp/rector-src/pull/7877

(it would be great if you could help me add this to other rules as well)

calebdw (approved):
approved after ci passes and changes are made 👍

peterfox (approved):

src/Rector/Class_/RouteKeyMethodToRouteKeyAttributeRector.php
MrPunyapal: Thanks @calebdw! I have started implementing this, but hit two blockers that make the test suite fail.

1. rector 2.5.9 (latest) crashes on any rule implementing this interface due to a bug in `InstalledPackageResolver`. It is fixed upstream in rectorphp/rector-src#8237 but not yet released.

2. Even after that fix, `laravel/framework` is not installed in this repo's dev env (we use stubs), so the rule gets filtered out and the transform fixtures fail. Existing implementations (rector-phpunit, rector-symfony) work because the constrained package is in their `require-dev`.

How would you like to handle this? Should I add `laravel/framework` to `require-dev` and wait for a rector release that includes the fix?

calebdw: It looks like rector has made 2 releases, can you try now and and see if that's enough?

MrPunyapal: Good news: the rector 2.6.1 release does fix the `InstalledPackageResolver` crash, so the interface no longer errors in tests.

One blocker remains though. Even with the fix, `laravel/framework` is not installed in this repo's dev env (we use stubs), so `resolvePackageVersion('laravel/framework')` returns `null` and the rule is filtered out. That makes the transform fixture fail (the diff is not applied), while the `skip_*` fixtures pass.

So the rector release alone is not enough. Do you want me to add `laravel/framework` to `require-dev` so the resolver finds it (like rector-phpunit/rector-symfony do with their constrained packages)?

calebdw: Hmm, so the problem is that while adding it to require-dev would work for this rule it would not work for other rules that might have a version constraint `<9.0` for example


Rector gets around this for php versions by setting this in the test config:
```php
return static function (RectorConfig $rectorConfig): void {
    $rectorConfig->rule(ReadOnlyAnonymousClassRector::class);

    $rectorConfig->phpVersion(PhpVersionFeature::READONLY_ANONYMOUS_CLASS);
};
```

Maybe we need to add something similar for composer packages (name and version) so that the rule will run in tests.

Any thoughts on what we can do here @TomasVotruba? Don't really want to have to include the framework as a dependency just so the tests will run

MrPunyapal: I found a solution that avoids adding the framework as a dependency.

Rector 2.6.1 ships a built-in hook for exactly this: `AbstractRectorTestCase::provideComposerJsonFilePath()` (marked `@api used by extensions`). It redirects the `InstalledPackageResolver` to read package versions from a standalone `composer.json` instead of `vendor/composer/installed.json`, so the rule can be tested against a package that is not installed.

```php
protected function provideComposerJsonFilePath(): ?string
{
    return __DIR__ . '/composer.json';
}
```

With a test-local `composer.json`:

```json
{
    "require": {
        "laravel/framework": "^13.21"
    }
}
```

The resolver reads the "require"/"require-dev" sections and uses the lowest version the constraint allows, so this also covers the `<9.0` case (just put `^8.0` in the test composer.json). I've implemented it on `RouteKeyMethodToRouteKeyAttributeRector` and all tests pass with no framework installed:

```
OK (643 tests, 830 assertions)
```

What are your thoughts on this approach?

calebdw: Ah yeah that should work!

calebdw:
Thanks!

src/Rector/Class_/RouteKeyMethodToRouteKeyAttributeRector.php
TomasVotruba: Yes, that's the way to go it :) custom composer.json in test 🎉

PR #540, opened 2026-08-05
feat(rules): add EmptyGuardedPropertyToUnguardedAttributeRector rule

COCOUVI:
## Overview

Adds a new Rector rule to migrate empty `$guarded` model properties to the Laravel 13 `#[Unguarded]` attribute.

### Example

```php
// Before
class User extends Model
{
    protected $guarded = [];
}

// After
#[Unguarded]
class User extends Model
{
}

### CODE REVIEW

File: src/Rector/Class_/EmptyGuardedPropertyToUnguardedAttributeRector.php [hunk 1]
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * @see EmptyGuardedPropertyToUnguardedAttributeRectorTest
+ */
+final class EmptyGuardedPropertyToUnguardedAttributeRector extends AbstractRector implements ComposerPackageConstraintInterface
+{
+    public function __construct(private readonly PhpAttributeAnalyzer $phpAttributeAnalyzer) {}
+
+    public function provideComposerPackageConstraint(): ComposerPackageConstraint
+    {
+        return new ComposerPackageConstraint('laravel/framework', '>=13.0');
```
calebdw, line 31: Did this unguarded attribute really come out in `13.0` or a later version?

src/Rector/Class_/EmptyGuardedPropertyToUnguardedAttributeRector.php
```diff
@@ -0,0 +1,112 @@
...
+            return null;
+        }
+
+        if ($propertyProperty->default->items !== []) {
+            return null;
+        }
+
+        if ($this->phpAttributeAnalyzer->hasPhpAttribute($node, 'Illuminate\Database\Eloquent\Attributes\Unguarded')) {
+            return null;
+        }
```
calebdw: this should probably come first right after the model check

docs/rector_rules_overview.md
calebdw, line 1: this file seems to have more changes than necessary...did your editor apply formatting rules?

COCOUVI:
Thanks for the review! Regarding the Laravel version constraint: the #[Unguarded] attribute was introduced in the initial Laravel 13.0 release, so I kept the >=13.0 constraint to stay consistent with the other Eloquent attribute rules (such as GuardedPropertyToGuardedAttributeRector).
I've also made the requested changes:
- Moved the hasPhpAttribute() check right after the model check to return early.
- Cleaned up the unrelated changes in rector_rules_overview.md. The extra diff was caused by trailing whitespace generated by the documentation generator, and the file now only contains the required rule documentation update.

calebdw:
What should happen if a model defines both a non-empty `$fillable` array and an empty `$guarded` array? Can you add a test for that case?

src/Rector/Class_/EmptyGuardedPropertyToUnguardedAttributeRector.php
```diff
@@ -109,4 +113,23 @@ public function refactor(Node $node): ?Node
 
         return $node;
     }
+
+    private function hasNonEmptyFillableProperty(Class_ $class): bool
+    {
+        $fillableProperty = $class->getProperty('fillable');
+        if ($fillableProperty === null) {
+            return false;
+        }
```
calebdw: this might be more complicated than necessary? You might just be able to check if hasProperty $fillable?

What if the fillable property and the guarded property are both empty? You'll have to check the framework to see which one wins...

COCOUVI:
Okay Sir. In this case, the rule should skip the conversion.An empty $guarded array with a non-empty $fillable array does not mean the model is fully unguarded. Laravel still relies on the $fillable definition to control mass assignment. Converting it to #[Unguarded] would remove that protection and change the model's behavior.
I've updated the rule to ignore this case and added a regression fixture to cover it:
tests/Rector/Class_/EmptyGuardedPropertyToUnguardedAttributeRector/Fixture/skip_with_fillable.php.inc. Test are passing ......

src/Rector/Class_/EmptyGuardedPropertyToUnguardedAttributeRector.php
COCOUVI: I checked Laravel's behavior and you were right, the extra $fillable check was unnecessary.
There is no precedence issue when both $fillable and $guarded are defined (including when both are empty). In Laravel 13, #[Unguarded] behaves the same as protected $guarded = [] and does not alter the $fillable behavior.
I removed the additional helper and added a regression test covering a model with a non-empty $fillable and an empty $guarded property.

calebdw:
Thanks!

PR #541, opened 2026-08-07
feat(rules): Add RelationTableStringToPivotClassRector rule

MizouziE:
## Rule

- Rewrites the `$table` argument of `belongsToMany()`, `morphToMany()` and `morphedByMany()` from a raw table-name string to the pivot model's class constant: `belongsToMany(Tag::class, 'post_tag')` -> `belongsToMany(Tag::class, PostTag::class)`.
- Handles the `$table` param wherever it sits; arg 1 for `belongsToMany`, arg 2 for the morph variants — including named arguments (`table: 'post_tag'`).

## Reasons

- Laravel already treats it as first-class: `BelongsToMany::resolveTableName()` accepts a model class, resolves it to `$model->getTable()`, and additionally calls `using($table)` when that model uses the `AsPivot` trait — so the class form carries strictly more information than the string.
- The table name stops being duplicated across the codebase. Rename the table in the pivot model's `#[Table]` attribute (or `$table`) and every relation follows.
- Static analysis and IDEs can see the reference: jump-to-definition, find-usages and automated class renames all work on `PostTag::class` and none of them work on `'post_tag'`.
- A typo in a class constant is a fatal error at parse/analysis time but a typo in the string is a runtime SQL error. Better to catch earlier than later.


src/Rector/MethodCall/RelationTableStringToPivotClassRector.php
```diff
@@ -0,0 +1,244 @@
...
+use Webmozart\Assert\Assert;
+
+/**
+ * @see RelationTableStringToPivotClassRectorTest
+ */
+final class RelationTableStringToPivotClassRector extends AbstractRector implements ConfigurableRectorInterface
+{
+    public const string MODEL_NAMESPACES = 'model_namespaces';
+
+    private const string MODEL_CLASS = 'Illuminate\Database\Eloquent\Model';
```
calebdw: pretty sure that this is already a constant elsewhere---and there's likely already a service to check if a class is a model

src/Rector/MethodCall/RelationTableStringToPivotClassRector.php
```diff
@@ -0,0 +1,244 @@
...
+
+    private const string TABLE_ATTRIBUTE_CLASS = 'Illuminate\Database\Eloquent\Attributes\Table';
```
calebdw: same thing, can we move this to a service or something, there's probably already duplicate constants...

src/Rector/MethodCall/RelationTableStringToPivotClassRector.php
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    private function getTableArgument(MethodCall $methodCall): ?Arg
+    {
+        foreach (self::TABLE_ARGUMENT_POSITIONS as $method => $position) {
+            if (! $this->isName($methodCall->name, $method)) {
+                continue;
+            }
+
+            foreach ($methodCall->getArgs() as $index => $arg) {
```
calebdw: you should directly be able to use `->getArg(name, position)` and clean this up

src/Rector/MethodCall/RelationTableStringToPivotClassRector.php
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        $table = $nativeReflection->getDefaultProperties()['table'] ?? null;
+
+        if (is_string($table)) {
+            return $table;
+        }
+
+        return $this->snake($this->pluralise($nativeReflection->getShortName()));
+    }
```
calebdw: This should be on a service class if not already. Pretty sure there's already one to find the table for a model

src/Rector/MethodCall/RelationTableStringToPivotClassRector.php
```diff
@@ -0,0 +1,244 @@
...
+        if (str_ends_with($lower, 'y') && preg_match('/[aeiou]y$/', $lower) === 0) {
+            return substr($word, 0, -1) . 'ies';
+        }
+
+        if (preg_match('/(s|x|z|ch|sh)$/', $lower) === 1) {
+            return $word . 'es';
+        }
+
+        return $word . 's';
+    }
```
calebdw: not really a fan of all these helpers---sounds like we just need to use Laravel Str

src/Rector/MethodCall/RelationTableStringToPivotClassRector.php
```diff
@@ -0,0 +1,244 @@
...
+use RectorLaravel\AbstractRector;
+use RectorLaravel\Tests\Rector\MethodCall\RelationTableStringToPivotClassRector\RelationTableStringToPivotClassRectorTest;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\ConfiguredCodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+use Webmozart\Assert\Assert;
+
+/**
+ * @see RelationTableStringToPivotClassRectorTest
+ */
+final class RelationTableStringToPivotClassRector extends AbstractRector implements ConfigurableRectorInterface
```
calebdw: please implement the composer version constraint interface

src/Rector/MethodCall/RelationTableStringToPivotClassRector.php
```diff
@@ -0,0 +1,244 @@
...
+            return null;
+        }
+
+        $table = $tableArgument->value;
+
+        if (! $table instanceof String_) {
+            return null;
+        }
+
+        $pivotClass = $this->resolvePivotClass($table->value);
```
calebdw: many times, if a pivot table is actual in use, then it's probably already in the method generics or in `->using()`---it would be good to support both of these

```
    /** @return BelongsToMany<User, $this, WorkOrderPersonnel> */
```

src/Rector/MethodCall/RelationTableStringToPivotClassRector.php
```diff
@@ -0,0 +1,244 @@
...
+                new ConfiguredCodeSample(
+                    <<<'CODE_SAMPLE'
+$this->belongsToMany(Tag::class, 'post_tag');
+CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
+$this->belongsToMany(Tag::class, \App\Models\PostTag::class);
+CODE_SAMPLE
+                    ,
+                    [RelationTableStringToPivotClassRector::MODEL_NAMESPACES => ['App\Models']]
```
calebdw: this is a bit brittle and not all that useful for folks that use DDD or modules where there might be a ton of different Model namespaces---any thoughts on how to fully support this without having to configure the rule?

calebdw (changes requested):

calebdw:
> Set registration: left out of every set deliberately; laravel-code-quality is the best fit if desired?

Yeah that's probably a good fit

MizouziE:
@calebdw Thank you very much for all the guidance here. I have taken it all on board and it led me to make a lot of improvements which hopefully satisfy the concerns raised. 

## Changes

Most notably the rule was lifted up from MethodCall to ClassMethod which allows it to now examine chained method nodes more easily and the inference from the string is demoted down to a fallback. It now goes `->using()` > doc-block generics > string inference.

### Helpers > `illuminate/support`

I was also able to remove the primitive helper methods and replace them with illuminate/support real versions. I made sure to check that versions of laravel as far back as the existing constraints would be compatible (the real constraint is the PHP version, so laravel 6) and the methods used have been stable for a very long time. This also allowed for the removal of the composer version constraint which was the correct call for the old shape of things, but now it was redundant due to the full backwards compatibility.

### Service Classes

Initial attempts to implement the existing service classes led me to adding the ability for ModelAnalyzer to also derive a table name from the Attribute. Further to that I ended up creating a new Node analyzer class with tests specifically for the Pivot models. Thanks to this we could drop the brittle configurable array for model namespaces.

Another notable change due to the ModelAnalyzer is that the Model stub no longer has the hard-coded fallback table name and instead mirrors Laravel closer seeing as the Str methods are now available to it. This only had one test to update, so I judged it as an acceptable change.

### Other

- I took your comment RE: sets as a green light for adding to the existing set
- I had a little confusion (now reverted in this branch) with running phpunit and hitting OOM errors, but that does not affect CI as it runs with `memory-limit=-1`
- Configuration is no longer required thanks to the much improved derivation of namespaces

src/NodeAnalyzer/ModelAnalyzer.php
```diff
@@ -32,6 +33,17 @@ protected static function relationType(): ObjectType
      */
     public function getTable(string|ObjectType $model): ?string
     {
+        if (! is_string($model)) {
+            /** @var class-string<Model> $model */
+            $model = $model->getClassName();
+        }
+
+        $table = $this->resolveTableFromAttribute($this->getClass($model));
```
calebdw, line 41: why can this just not flow through the path below? `->getTable()` should return the correct table even if using the attribute

stubs/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
```diff
@@ -6,4 +6,22 @@
...
+ */
+class BelongsToMany extends Relation
+{
+    /**
+     * Specify the custom pivot model to use for the relationship.
+     *
+     * @param  class-string  $class
+     * @return $this
+     */
+    public function using($class) {}
```
calebdw, line 26: this is not the correct generics for the `using` method---check the framework, pretty sure there's a this-out doc

composer.json
```diff
@@ -5,6 +5,7 @@
     "description": "Rector upgrades rules for Laravel Framework",
     "require": {
         "php": ">=8.3",
+        "illuminate/support": "^6.0 || ^7.0 || ^8.0 || ^9.0 || ^10.0 || ^11.0 || ^12.0 || ^13.0",
```
calebdw, line 8: not really a fan of adding this here:
- should be added to `require-dev` so that this works in testing (can just have the latest version)
- should be listed under the `suggest` key with the appropriate supported version (e.g., 6-13)

this is similar to how the framework operates:

Image

src/Rector/ClassMethod/RelationTableStringToPivotClassRector.php
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * @param  ClassMethod  $node
+     */
+    public function refactor(Node $node): ?ClassMethod
+    {
+        $hasChanged = false;
+
+        foreach ($this->betterNodeFinder->findInstancesOfInFunctionLikeScoped($node, MethodCall::class) as $methodCall) {
+            $tableArgument = $this->pivotClassAnalyzer->matchRelationTableArg($methodCall);
```
calebdw, line 72: the rector convention is to define a `shouldSkip()` method that gets checked up front before processing the rest of the rector. As it is, this rule is very inefficient:
- should immediately check if the method is in a model class, if not skip
- should immediately check the native return type, if not one of the 3 relations that matter then skip
- etc.



```suggestion
    /**
     * @param  ClassMethod  $node
     */
    public function refactor(Node $node): ?ClassMethod
    {
        if ($this->shouldSkip($node)) {
            return null;
        }
        
        $hasChanged = false;

        foreach ($this->betterNodeFinder->findInstancesOfInFunctionLikeScoped($node, MethodCall::class) as $methodCall) {
            $tableArgument = $this->pivotClassAnalyzer->matchRelationTableArg($methodCall);
```

MizouziE:
@calebdw, thanks once again for the pointers. I think I have addressed all concerns and want to explain one or two discoveries along the way.

## Changes

- `->getTable()` : Now working properly as it goes through the framework and resolves automatically with the Table attribute available in v13, which gives the nice side effect of enabling `getPrimaryKey()` to resolve from `#[Table(key:...)]`. By making some improvements to the stub, it now has a constructor calling `initializeModelAttributes()`, it's a closer reflection of the framework.
- `using()` generics: Fixed and now align properly with the exception of being bound to Model not Pivot, but the doc-block explains why.
- require-dev + suggest: Done, much cleaner and also helped me realise that `AsPivot` was introduced in framework v8, so the list is slightly shorter than before, which shores things up a bit.
- `shouldSkip()`: added and purposely allows traits through, though in a trait `$this` is the trait, so the joining-table convention can't be verified and a `belongsToMany` there resolves only via `using()`, generics or an `AsPivot` model
- phpunit.xml memory_limit 512M: I have upped the memory limit for local development PHPUnit runs as the loading of the illuminate/support package tips it over the edge of the default 128M. I measured 122.5MB on main vs 192.5MB here and I believe it is down to PHPStan reflecting real `Illuminate` classes rather than stubs.

PR #543, opened 2026-08-20
Bump Rector to ^2.6.3 and migrate SetProvider to composer based

samsonasik:
Fix phpstan notice:

```bash
  Line   src/Set/LaravelSetProvider.php                                         
 ------ ----------------------------------------------------------------------- 
  14     Class RectorLaravel\Set\LaravelSetProvider implements deprecated       
         interface Rector\Set\Contract\SetProviderInterface:                    
         Bond the rules themselves instead, by implementing the                 
         ComposerPackageConstraintInterface. A set                              
         described as an object only existed to be matched against the          
         installed packages; a bonded rule states the exact                     
         package version its target API is available from and applies from      
         there upwards, so a plain set file is enough.                          
```

shown at https://github.com/driftingly/rector-laravel/actions/runs/32388778123/job/96489605157?pr=542#step:5:178

The `SetProviderInterface` can be removed in the future as already deprecated. This PR migrate to composer based.


samsonasik:
@peterfox @calebdw  ready to review/merge 👍

PR #545, opened 2026-08-25
Move composer-based set to single config with version-bound rule registration

TomasVotruba:
Following [palantirnet/drupal-rector#419](https://github.com/palantirnet/drupal-rector/pull/419), this moves the composer-based set from a version-gated **set-import loop** to a **single config that registers each rule directly**, bound to the package version that introduced it.

## Before

A `SetProvider` listed whole per-minor set files and composer-triggered sets:

```php
final class LaravelSetProvider implements SetProviderInterface
{
    public function provide(): array
    {
        return [
            // ...
            new ComposerTriggeredSet('laravel', 'fakerphp/faker', '1.0', FakerSetList::FAKER_10),
            new ComposerTriggeredSet('laravel', 'livewire/livewire', '3.0', LivewireSetList::LIVEWIRE_30),
            ...$this->getLaravelVersions(), // LARAVEL_50 ... LARAVEL_130
        ];
    }
}
```

## After

Every rule is registered once, bound to its introducing version:

```php
// configurable (core) rules
$rectorConfig->ruleWithConfigurationComposerVersionBound(RenameMethodRector::class, [
    new MethodCallRename('Illuminate\Http\Request', 'get', 'input'),
], 'laravel/framework', '>=12.0');

// non-configurable rules self-filter via ComposerPackageConstraintInterface
$rectorConfig->rule(ContainerBindConcreteWithClosureOnlyRector::class);
```

The 25 remaining non-configurable `RectorLaravel` rules now implement `ComposerPackageConstraintInterface`, matching the 34 attribute rules that already used it:

```diff
-final class Redirect301ToPermanentRedirectRector extends AbstractRector
+final class Redirect301ToPermanentRedirectRector extends AbstractRector implements ComposerPackageConstraintInterface
 {
+    public function provideComposerPackageConstraint(): ComposerPackageConstraint
+    {
+        return new ComposerPackageConstraint('laravel/framework', '>=5.7');
+    }
```

A rule stays active from its version upwards, so a direct upgrade across several majors is covered by this one config.

## Tests

`laravel/framework` isn't installed in this repo, so a version-bound rule is otherwise filtered out of its own fixture run. Each converted rule test now ships a `composer.json` fixture and overrides `provideComposerJsonFilePath()`, following the convention already established by the attribute rules:

```php
protected function provideComposerJsonFilePath(): string
{
    return __DIR__ . '/composer.json';
}
```


TomasVotruba:
Ready for review :+1: 

tl;dr;

* Rector moved from compsoer-based sets that had to be grouped in many-tiny-hard-to-maintain-rule lists, to package+version bound rules - https://getrector.com/documentation/composer-based-sets#content-make-a-rule-bound-to-a-package-version
* each rule is triggered by exact package + version it was added in :+1: 
* the `ApplicationAnalyzer` that had same job, was removed and replacted by the same logic
* to add a new rule, use `composer-based.php` set - this file is picked by Rector automatically

calebdw (approved):
look good to me, thanks!

PR #546, opened 2026-08-25
Move Cashier to composer-based set, deprecate Cashier + Livewire set lists

TomasVotruba:
Smaller brother to #545, now the rest of packages are also handled by composer version too.

Extends the composer-based set to Cashier, and deprecates the remaining hand-maintained package version set lists now covered by `withComposerBased(laravel: true)`.

Let me know if you'll come across any trigger issues. Will make sure to handle those before release :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,11 @@ return RectorConfig::configure()
     ->withComposerBased(laravel: true, /** other options */);
 ```
 
-This registers `RectorLaravel\Set\LaravelSetList::COMPOSER_BASED`, which applies the version sets of Laravel (and of Faker and Livewire, when installed) up to the installed version.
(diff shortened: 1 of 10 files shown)


calebdw:
Thanks! Sure do love the new way of doing things

TomasVotruba:
Awesome, me even more :smile: 

You've nailed a groundbreaking feature in there: https://github.com/rectorphp/rector-src/pull/7877 :clap:

calebdw:
You know---we could also use the same concept but for PHP versions (currently we only have the min php version interface)

TomasVotruba:
That could also work. Which PHP features would benefit from an upper limit though?

calebdw:
potentially downgrade rules, or maybe there's a php upgrade rule that needs a min and max bound

PR #549, opened 2026-08-26
Add AssertDatabaseCountToAssertDatabaseEmptyRector

MizouziE:
## Rule

This adds a rule to convert zero count `assertDatabaseCount()` calls to `assertDatabaseEmpty()`.

Example:
```diff
 class ExampleTest extends \Illuminate\Foundation\Testing\TestCase
 {
     public function testFoo()
     {
-        $this->assertDatabaseCount('users', 0);
-        $this->assertDatabaseCount(User::class, 0);
-        $this->assertDatabaseCount('users', 0, 'other');
+        $this->assertDatabaseEmpty('users');
+        $this->assertDatabaseEmpty(User::class);
+        $this->assertDatabaseEmpty('users', 'other');
     }
 }
```

## Reason

- Since Laravel 9.39 we've had the more readable empty database table assertions available to us
- We have noticed lately that coding agents repeatedly use zero counts in database assertions


src/Rector/MethodCall/AssertDatabaseCountToAssertDatabaseEmptyRector.php
```diff
@@ -0,0 +1,117 @@
...
+        if (count($node->getRawArgs()) > 3) {
+            return null;
+        }
+
+        // getArg() skips unpacked args, so a spread would resolve to null and be dropped silently
+        foreach ($node->getRawArgs() as $rawArg) {
+            if ($rawArg instanceof Arg && $rawArg->unpack) {
+                return null;
+            }
+        }
```
calebdw, line 94: there's no need for this---the `getArg` below handles this case

MizouziE: Aha, yes you're correct. I see that a similar problem was solved recently in `ContainerBindConcreteWithClosureOnlyRector` so I will borrow that shape to fix here too. Thanks!

PR #559, opened 2026-09-03
Add UnlinkFuncCallToFileFacadeDeleteRector

chinmaypurav:
Adds a rule that replaces raw `unlink()` calls with the `File` facade's `delete()` method, so applications go through the framework filesystem API rather than the bare PHP function.

```diff
-unlink($path);
+\Illuminate\Support\Facades\File::delete($path);
```

Registered in the `laravel-code-quality` set.

## Scope

`File::delete()` is a close wrapper around `@unlink()` and returns `bool`, the same as `unlink()`, so the rule refactors in any expression position rather than statement position only. `if (! unlink($path))` and `return unlink($path);` both convert safely.

First class callables convert too, so `array_map(unlink(...), $paths)` becomes `array_map(\Illuminate\Support\Facades\File::delete(...), $paths)`.

The error suppressor is **not** carried over:

```diff
-@unlink($path);
+\Illuminate\Support\Facades\File::delete($path);
```

`File::delete()` calls `@unlink()` internally and catches the resulting `ErrorException`, so no warning escapes and the `@` is redundant. This is why `ErrorSuppress` is one of the rule's node types — returning a node from the inner `FuncCall` alone would replace only the call and leave the `@` behind.

Two cases are skipped, where the swap would not be equivalent:

- **`unlink($path, $context)`** — the stream context argument has no facade equivalent, so the two argument form is left alone rather than silently dropping the context.
- **`unlink(...$args)`** — an unpacked argument may itself carry that stream context.

A `filename:` named argument is converted to a positional one, since the facade parameter is named `$paths`.

## Caveats for reviewers

**Warning behaviour changes.** `unlink()` emits a PHP warning on failure, whereas `File::delete()` suppresses it internally and returns `false`. Code relying on an error handler to convert that warning into an `ErrorException` will stop throwing after this rewrite, and will need to check the return value instead. This is inherent to the target API rather than to the rule, but it is a real behavioural difference and the reason the rule lives in a code quality set rather than an upgrade set.

**`File` resolves through the container.** The facade needs a booted application, so `unlink()` calls in code that runs outside the framework lifecycle are worth reviewing in the `--dry-run` diff before applying.

**Callable arity differs.** `unlink(...)` accepts an optional second stream context argument that `File::delete(...)` does not. A converted callable invoked with two arguments would behave differently, though passing a stream context through a callable reference is an unlikely shape in practice.

## Tests

Six fixtures covering the basic conversion across expression positions, error suppressed calls, first class callables, the named argument form, and the two skip cases. `composer rector`, `composer lint`, `composer phpstan`, `composer docs`, and `composer structarmed` are all clean.

Note that the suite has 8 pre-existing failures on `main` in the `ArrayDimFetch` and `PropertyFetch` rules, unrelated to this change. They reproduce on a clean checkout and are untouched here.

### DIFF
--- a/src/Rector/FuncCall/UnlinkFuncCallToFileFacadeDeleteRector.php
+++ b/src/Rector/FuncCall/UnlinkFuncCallToFileFacadeDeleteRector.php
@@ -6,9 +6,11 @@ namespace RectorLaravel\Rector\FuncCall;
 
 use PhpParser\Node;
 use PhpParser\Node\Arg;
+use PhpParser\Node\Expr\ErrorSuppress;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Expr\StaticCall;
-use PhpParser\Node\Identifier;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\VariadicPlaceholder;
 use RectorLaravel\AbstractRector;
 use RectorLaravel\Tests\Rector\FuncCall\UnlinkFuncCallToFileFacadeDeleteRector\UnlinkFuncCallToFileFacadeDeleteRectorTest;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
@@ -45,29 +47,49 @@ CODE_SAMPLE
      */
     public function getNodeTypes(): array
     {
-        return [FuncCall::class];
+        return [ErrorSuppress::class, FuncCall::class];
     }
 
     /**
-     * @param  FuncCall  $node
+     * @param  ErrorSuppress|FuncCall  $node
      */
     public function refactor(Node $node): ?StaticCall
     {
-        if (! $this->isName($node->name, 'unlink')) {
+        // File::delete() suppresses the failure itself, so the suppressor is dropped
+        $funcCall = $node instanceof ErrorSuppress ? $node->expr : $node;
+
+        if (! $funcCall instanceof FuncCall) {
             return null;
+        }
+
+        if (! $this->isName($funcCall->name, 'unlink')) {
+            return null;
+        }
+
+        if ($funcCall->isFirstClassCallable()) {
+            return $this->createFileDeleteCall([new VariadicPlaceholder]);
         }
 
         // the stream context argument has no equivalent on the facade
-        if (count($node->args) !== 1) {
+        if (count($funcCall->args) !== 1) {
             return null;
         }
 
-        $arg = $node->getArg('filename', 0);
+        $arg = $funcCall->getArg('filename', 0);
 
         if (! $arg instanceof Arg) {
             return null;
         }
 
-        return $this->nodeFactory->createStaticCall('Illuminate\Support\Facades\File', 'delete', [$arg->value]);
+        // the facade parameter is named $paths, so a filename: argument is passed positionally
+        return $this->createFileDeleteCall([new Arg($arg->value)]);
+    }
+
+    /**
+     * @param  array<Arg|VariadicPlaceholder>  $args
+     */
+    private function createFileDeleteCall(array $args): StaticCall
+    {
+        return new StaticCall(new FullyQualified('Illuminate\Support\Facades\File'), 'delete', $args);
     }
 }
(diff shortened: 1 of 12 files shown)


src/Rector/FuncCall/UnlinkFuncCallToFileFacadeDeleteRector.php
```diff
@@ -0,0 +1,82 @@
...
+        if ($arg->name instanceof Identifier && ! $this->isName($arg->name, 'filename')) {
+            return null;
+        }
+
+        return $this->nodeFactory->createStaticCall(
+            'Illuminate\Support\Facades\File',
+            'delete',
+            [$arg->value],
+        );
+    }
```
calebdw: this is all that's needed

```suggestion
        $arg = $node->getArg('filename', 0);

        if (! $arg instanceof Arg) {
            return null;
        }

        return $this->nodeFactory->createStaticCall('Illuminate\Support\Facades\File', 'delete', [$arg->value]);
    }
```

tests/Rector/FuncCall/UnlinkFuncCallToFileFacadeDeleteRector/Fixture/error_suppressed.php.inc
```diff
@@ -0,0 +1,27 @@
...
+-----
+<?php
+
+namespace RectorLaravel\Tests\Rector\FuncCall\UnlinkFuncCallToFileFacadeDeleteRector\Fixture;
+
+class ErrorSuppressed
+{
+    public function run(string $path)
+    {
+        @\Illuminate\Support\Facades\File::delete($path);
```
calebdw: the `@` suppressor should NOT be carried over---the File::delete does not emit a warning so this is not needed

tests/Rector/FuncCall/UnlinkFuncCallToFileFacadeDeleteRector/Fixture/skip_unpacked_and_first_class_callable.php.inc
```diff
@@ -0,0 +1,15 @@
...
+
+namespace RectorLaravel\Tests\Rector\FuncCall\UnlinkFuncCallToFileFacadeDeleteRector\Fixture;
+
+class SkipUnpackedAndFirstClassCallable
+{
+    public function run(array $arguments, array $paths)
+    {
+        unlink(...$arguments);
+
+        array_map(unlink(...), $paths);
```
calebdw: there should be no reason why we also can't support `File::delete(...)`

tests/Rector/FuncCall/UnlinkFuncCallToFileFacadeDeleteRector/Fixture/error_suppressed.php.inc
chinmaypurav: Agreed, and fixed in 83bcdd9. `File::delete()` calls `@unlink()` internally and catches the `ErrorException`, so nothing surfaces and the suppressor is dead weight.

Doing this needed `ErrorSuppress` added to `getNodeTypes()` — returning a node from the inner `FuncCall` can only replace the call itself, leaving the `@` in place. Handling the outer node replaces the whole expression.

tests/Rector/FuncCall/UnlinkFuncCallToFileFacadeDeleteRector/Fixture/skip_unpacked_and_first_class_callable.php.inc
chinmaypurav: Done in 83bcdd9 — `unlink(...)` now converts to `\Illuminate\Support\Facades\File::delete(...)`.

Note `createStaticCall()` cannot build this one: it routes args through `BuilderFactory::args()`, which wraps any non-`Arg` in `new Arg(BuilderHelpers::normalizeValue(...))` and throws on a `VariadicPlaceholder`. So the `StaticCall` is constructed directly.

Unpacked args stay skipped, since `unlink(...$args)` may carry the stream context. That falls out of `getArg()` for free — it skips `unpack` args and returns null for first class callables. Split the old fixture into `first_class_callable.php.inc` and `skip_unpacked_argument.php.inc`.

PR #560, opened 2026-09-03
Add CommandPropertyToAsCommandAttributeRector

### LINKED ISSUE #383: Add AsCommand rector
Similar to Symfony rectors

```php
        \Rector\Symfony\Symfony61\Rector\Class_\CommandPropertyToAttributeRector::class,
        \Rector\Symfony\Symfony61\Rector\Class_\CommandConfigureToAttributeRector::class,
```

- https://getrector.com/rule-detail/command-property-to-attribute-rector
- https://getrector.com/rule-detail/command-configure-to-attribute-rector

Symfony commands have:

```php
    protected $signature = '...';
    protected $description = '...';
```

chinmaypurav:
Closes #383.

Moves a console command's `$signature`/`$name` and `$description` into Symfony's `#[AsCommand]`.

```diff
 use Illuminate\Console\Command;
+use Symfony\Component\Console\Attribute\AsCommand;

+#[AsCommand(name: 'mail:send', description: 'Send the queued emails')]
 class SendEmails extends Command
 {
-    protected $signature = 'mail:send';
-
-    protected $description = 'Send the queued emails';
 }
```

### Why this is behaviour preserving

Symfony's `Command` declares `$name`, `$description`, `$help` and `$hidden` as **private**, so Laravel's `protected` redeclarations occupy separate slots. Symfony's constructor reads the attribute into its own slots, and Laravel's constructor only overwrites them when its own properties are non-empty. Removing the properties therefore hands control to the attribute — for `$signature`, `$name` and `$description`.

`$help`, `$hidden` and `$aliases` are deliberately left in place: the `help` argument needs Symfony Console 7.3, and Laravel's constructor unconditionally overwrites help and hidden from its own properties, so moving those would be a behaviour change.

### The signature problem

As raised in the issue, a signature carrying arguments or options can't be expressed as an attribute name, since Laravel parses the input definition out of it. Those are skipped.

Related, and easy to miss: dropping `$signature` makes Laravel call `specifyParameters()`, which reactivates `getArguments()`/`getOptions()`. The rule skips classes that override either, checked through the inheritance chain via reflection rather than just the class body.

### Other cases skipped

Signatures with whitespace or `|` alias pipes; classes declaring both `$signature` and `$name` (the signature silently wins, so that's a human's call); abstract and anonymous classes; an existing `#[AsCommand]`; non-`protected`, `static`, `readonly` or grouped (`$a = 'x', $b = 'y'`) properties; non-literal defaults; empty signatures; and classes reading `$this->signature`/`$this->description` in their own body.

### Cases handled rather than skipped

`$description = ''` — property dropped, argument omitted, since that's the default anyway. A non-literal `$description` next to a literal signature — the name is converted and the description property stays put. `$name` alongside `getArguments()` — safe, because `specifyParameters()` already runs on that path. Existing attribute groups are appended to.

### Notes

Gated on `laravel/framework >=9.0` (`AsCommand` needs Symfony Console 6+) and `PhpVersionFeature::ATTRIBUTES`.

I left the rule out of every set on purpose — it isn't tied to a Laravel version bump, and it would collide with the Laravel 13 `#[Signature]`/`#[Description]` rules if both ran. Happy to add it wherever you'd prefer.

26 fixtures (8 transforming, 18 skip). PHPStan max, duster and structarmed are clean, docs regenerated. The suite has 8 failures on `main` already, all in `ArrayDimFetch`/`ReplaceFakerInstanceWithHelperRector` and unrelated to this change.


calebdw:
I like this, but this should probably handle when the arguments and options are specified in the signature string---it needs to handle properly creating the get{Options,Arguments} methods with the correctly instantiated classes

chinmaypurav:
> I like this, but this should probably handle when the arguments and options are specified in the signature string---it needs to handle properly creating the get{Options,Arguments} methods with the correctly instantiated classes

@calebdw the changes have been implemented. Let me know what you think.

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_grouped_properties.php.inc
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace RectorLaravel\Tests\Rector\Class_\CommandPropertyToAsCommandAttributeRector\Fixture;
+
+use Illuminate\Console\Command;
+
+class SkipGroupedProperties extends Command
+{
+    protected $signature = 'mail:send', $description = 'Send the queued emails';
```
calebdw, line 9: why is this skipped? we should be able to handle this no?

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_existing_as_command.php.inc
```diff
@@ -0,0 +1,16 @@
...
+
+use Illuminate\Console\Command;
+use Symfony\Component\Console\Attribute\AsCommand;
+
+#[AsCommand(name: 'mail:send')]
+class SkipExistingAsCommand extends Command
+{
+    protected $signature = 'mail:send';
+
+    protected $description = 'Send the queued emails';
```
calebdw, line 13: in this case I would expect a few things:
- this rule (or maybe a separate rule) that removes the unneeded properties
- I would expect the description to be moved into the `AsCommand`

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_inherited_parameter_methods.php.inc
```diff
@@ -0,0 +1,20 @@
...
+    protected function getArguments(): array
+    {
+        return [['user', null, 'The user']];
+    }
+}
+
+class SkipInheritedParameterMethods extends InheritedParameterMethodsCommand
+{
+    protected $signature = 'mail:send';
+}
```
calebdw, line 18: I don't see a reason to skip this?

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_non_string_signature.php.inc
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace RectorLaravel\Tests\Rector\Class_\CommandPropertyToAsCommandAttributeRector\Fixture;
+
+use Illuminate\Console\Command;
+
+class SkipNonStringSignature extends Command
+{
+    protected $signature = 'mail:' . 'send';
```
calebdw, line 9: no reason to skip this, this is a constant expr that can be parsed just fine?

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_property_used_in_class.php.inc
```diff
@@ -0,0 +1,19 @@
...
+
+class SkipPropertyUsedInClass extends Command
+{
+    protected $signature = 'mail:send';
+
+    protected $description = 'Send the queued emails';
+
+    public function handle(): void
+    {
+        echo $this->description;
```
calebdw, line 15: there is no reason to skip this---the Command constructor sets the `$description` property from the AsCommand description

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_signature_with_brace_in_name.php.inc
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace RectorLaravel\Tests\Rector\Class_\CommandPropertyToAsCommandAttributeRector\Fixture;
+
+use Illuminate\Console\Command;
+
+class SkipSignatureWithBraceInName extends Command
+{
+    protected $signature = 'mail:send{user}';
```
calebdw, line 9: is this even a possible command? We should only have fixtures for realistic examples?

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_signature_with_parameter_methods.php.inc
```diff
@@ -0,0 +1,17 @@
...
+use Illuminate\Console\Command;
+
+class SkipSignatureWithParameterMethods extends Command
+{
+    protected $signature = 'mail:send {user}';
+
+    protected function getOptions(): array
+    {
+        return [['queue', null, null, 'Queue the mail']];
+    }
```
calebdw, line 14: we should be able to add to the return array instead of just skipping altogether; also, you would be adding a `getArgument` method and not modifying the `getOptions` (but should have a test for that as well)

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_signature_with_stray_text.php.inc
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace RectorLaravel\Tests\Rector\Class_\CommandPropertyToAsCommandAttributeRector\Fixture;
+
+use Illuminate\Console\Command;
+
+class SkipSignatureWithStrayText extends Command
+{
+    protected $signature = 'mail:send {user} unparsed';
```
calebdw, line 9: is this realistic? If not then no reason to skip

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_signature_and_name.php.inc
```diff
@@ -0,0 +1,14 @@
...
+namespace RectorLaravel\Tests\Rector\Class_\CommandPropertyToAsCommandAttributeRector\Fixture;
+
+use Illuminate\Console\Command;
+
+class SkipSignatureAndName extends Command
+{
+    protected $signature = 'mail:send';
+
+    protected $name = 'mail:dispatch';
+}
```
calebdw, line 12: I don't see a reason to skip this---figure out which one has precedence in laravel, make the update, then remove the unneeded extra property?

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_static_property.php.inc
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace RectorLaravel\Tests\Rector\Class_\CommandPropertyToAsCommandAttributeRector\Fixture;
+
+use Illuminate\Console\Command;
+
+class SkipStaticProperty extends Command
+{
+    protected static $signature = 'mail:send';
```
calebdw, line 9: not sure that we need to skip this?

tests/Rector/Class_/CommandPropertyToAsCommandAttributeRector/Fixture/skip_signature_with_stray_text.php.inc
chinmaypurav: Agree. removed

PR #566, opened 2026-09-07
Resolve container binding arguments by name in ContainerBindConcreteWithClosureOnlyRector

### LINKED ISSUE #397: ContainerBindConcreteWithClosureOnlyRector Rule creates invalid code with named arguments
The rector rule `\RectorLaravel\Rector\MethodCall\ContainerBindConcreteWithClosureOnlyRector::class`, does not take into account if a singleton is called with named arguments:

This is what the rule currently outputs if its applied on the code:

```diff
        $this->app->singleton(
-            abstract: SomeClass::class,
-            concrete: function (Application $app, array $parameters) {
+            concrete: function (Application $app, array $parameters): SomeClass::class {
                // ...
            },
        );
```

Expected output:

```diff
        $this->app->singleton(
-           abstract: SomeClass::class,
-           concrete: function (Application $app, array $parameters) {
+           function (Application $app, array $parameters): SomeClass::class {
                // ...
            },
        );
```

Alternative expected output:

```diff
        $this->app->singleton(
-           abstract: SomeClass::class,
-           concrete: function (Application $app, array $parameters) {
+           abstract: function (Application $app, array $parameters): SomeClass::class {
                // ...
            },
        );
```

chinmaypurav:
Fixes #397.

## Named arguments produced fatal code

The rule read the abstract and concrete arguments positionally and then dropped index 0, so a call using named arguments lost `$abstract`:

```diff
 $this->app->singleton(
-    abstract: SomeClass::class,
-    concrete: function (Application $app, array $parameters) {
+    concrete: function (Application $app, array $parameters): SomeClass {
         // ...
     },
 );
```

`Container::singleton($abstract, $concrete = null)` still requires `$abstract`, so the result is an `ArgumentCountError`.

Arguments are now matched to parameters by `Arg::$name` when present and by position otherwise. That also picks up the reversed form, `singleton(concrete: $fn, abstract: SomeClass::class)`, which the old positional read skipped entirely. The closure has its name stripped when it moves into the abstract position, so the output is plain `singleton($fn)`.

## `$shared` was silently dropped

Not part of the issue, but found while writing fixtures for it. `bind(SomeClass::class, $fn, true)` became `bind($fn, true)`. Laravel routes a closure abstract through `bindBasedOnClosureReturnTypes($abstract, $concrete, $shared)`, which overwrites `$concrete` with the closure — so the `true` landed in `$concrete` and was discarded, quietly turning a shared binding into a non-shared one. It is now emitted as `bind($fn, shared: true)`.

## `bindIf()` / `singletonIf()` removed from the rule

Also outside the issue, and the largest behavioral change here. Both forward the abstract to `bound()`, which does `isset($this->bindings[$abstract])`. With a closure abstract that throws `TypeError: Cannot access offset of type Closure in isset or empty` — neither method has the closure handling that `bind()` gained. So every transform this rule made on `bindIf`/`singletonIf` produced code that fatals at boot, named arguments or not. Happy to split this into its own PR if you'd rather review it separately.

## Minor

Removed the `$classString instanceof Const_` guard. `Arg::$value` is always an `Expr` and `Const_` extends `NodeAbstract`, so it could never be true; PHPStan flags it as `instanceof.alwaysFalse` once the argument type is narrowed. No behavior change.

## Tests

New fixtures: `fixture_named_arguments`, `fixture_shared_argument`, `skip_conditional_binding`, `skip_spread_arguments`.

PHPStan (level max), `duster lint`, `rector --dry-run`, and `structarmed` are clean. The full suite goes 673 → 677 tests with the same 1 error / 14 failures that already fail on `main`, all in unrelated rules.

### DIFF
--- a/src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
+++ b/src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
@@ -77,13 +77,6 @@ CODE_SAMPLE
             return null;
         }
 
-        foreach ($node->getArgs() as $arg) {
-            // an unpacked argument hides which parameter each value lands in
-            if ($arg->unpack) {
-                return null;
-            }
-        }
-
         $abstractArg = $node->getArg('abstract', 0);
         $concreteArg = $node->getArg('concrete', 1);
         // singleton() has no $shared parameter, and passing one would be fatal
@@ -92,6 +85,12 @@ CODE_SAMPLE
             : null;
 
         if (! $abstractArg instanceof Arg || ! $concreteArg instanceof Arg) {
+            return null;
+        }
+
+        // getArg() cannot match a spread, and refuses a name that is not a parameter;
+        // rebuilding the call below would drop whatever it left behind
+        if (count(array_filter([$abstractArg, $concreteArg, $sharedArg])) !== count($node->getArgs())) {
             return null;
         }
 
--- a/src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
+++ b/src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
@@ -3,10 +3,11 @@
 namespace RectorLaravel\Rector\MethodCall;
 
 use PhpParser\Node;
-use PhpParser\Node\Const_;
+use PhpParser\Node\Arg;
 use PhpParser\Node\Expr\Closure;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Identifier;
 use PHPStan\Type\ObjectType;
 use Rector\PHPStanStaticTypeMapper\Enum\TypeKind;
 use Rector\StaticTypeMapper\StaticTypeMapper;
@@ -19,6 +20,18 @@ use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 
 final class ContainerBindConcreteWithClosureOnlyRector extends AbstractRector implements ComposerPackageConstraintInterface
 {
+    /**
+     * bindIf() and singletonIf() are left out on purpose: they pass the abstract
+     * straight to bound(), which uses it as an array offset, so a closure abstract
+     * fatals there.
+     *
+     * @var array<string, list<string>>
+     */
+    private const array PARAMETER_NAMES_BY_METHOD = [
+        'bind' => ['abstract', 'concrete', 'shared'],
+        'singleton' => ['abstract', 'concrete'],
+    ];
+
     public function __construct(
         private readonly ReturnTypeInferer $returnTypeInferer,
         private readonly StaticTypeMapper $staticTypeMapper,
@@ -61,7 +74,7 @@ CODE_SAMPLE
      */
     public function refactor(Node $node): ?MethodCall
     {
-        if (! $this->isNames($node->name, ['bind', 'singleton', 'bindIf', 'singletonIf'])) {
+        if (! $this->isNames($node->name, array_keys(self::PARAMETER_NAMES_BY_METHOD))) {
             return null;
         }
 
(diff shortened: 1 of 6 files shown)


src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
```diff
@@ -73,13 +86,18 @@ public function refactor(Node $node): ?MethodCall
...
             return null;
         }
 
-        $type = $this->getType($node->getArgs()[0]->value);
-        $classString = $node->getArgs()[0]->value;
-        $concreteNode = $node->getArgs()[1]->value;
+        [$abstractArg, $concreteArg, $sharedArg] = $arguments;
+
+        $classString = $abstractArg->value;
+        $concreteNode = $concreteArg->value;
```
calebdw, line 98: no need for all of this and the custom method---just use this

```php
$abstractArg = $node->getArg('abstract', 0);
$concreteArg = $node->getArg('concrete', 0);
$sharedArg = $node->getArg('shared', 0);
```

chinmaypurav: Thanks, done in 37e7fd4 — `getArg()` does exactly what the custom method was doing, and it's available since php-parser 5.6 which is what the `^5.6` constraint here allows.

Used positions 0/1/2 rather than 0/0/0, and kept two guards around it:

- `getArg()` `continue`s past unpacked args rather than reporting them, so `bind(Foo::class, $fn, ...$rest)` would resolve abstract and concrete and silently drop `...$rest`. Unpacked calls are now refused up front.
- `$shared` is only looked up for `bind()`. `singleton()` has no such parameter, and extra positional args to a userland method are legal and inert — so `singleton(Foo::class, $fn, true)` would have been rewritten to `singleton($fn, shared: true)`, turning working code into an `Unknown named parameter $shared` fatal. Covered in `fixture_shared_argument.php.inc`.

src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
```diff
@@ -73,13 +77,28 @@ public function refactor(Node $node): ?MethodCall
             return null;
         }
 
-        if (count($node->getArgs()) < 2) {
+        foreach ($node->getArgs() as $arg) {
+            // an unpacked argument hides which parameter each value lands in
+            if ($arg->unpack) {
+                return null;
+            }
+        }
```
calebdw: there is no need to check for this---`getArg` already properly handles this behind the scene

src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
```diff
@@ -109,9 +123,18 @@ public function refactor(Node $node): ?MethodCall
...
 
-        $args = $node->getArgs();
+        // the closure takes over the abstract position, so it must not stay named
+        $concreteArg->name = null;
+
+        $args = [$concreteArg];
+
+        if ($sharedArg instanceof Arg) {
+            // shared is no longer the third argument, so it has to be passed by name
+            $sharedArg->name = new Identifier('shared');
```
calebdw: should probably null coalesce here since it could already have a name

```suggestion
            $sharedArg->name ??= new Identifier('shared');
```

src/Rector/MethodCall/ContainerBindConcreteWithClosureOnlyRector.php
chinmaypurav: `getArg()` does match correctly — it never mis-assigns a spread to a parameter. But this rule rebuilds `$node->args` from the resolved arguments, so anything `getArg()` skipped disappears from the call. I removed the check to confirm:

```diff
-$app->bind(SomeInterface::class, function () {
+$app->bind(function (): SomeInterface {
     return new SomeClass();
-}, ...$rest);
+});
```

You're right that the loop was the wrong shape though. Replaced it in 6f4c4d2 with the invariant it was actually protecting — every argument has to be accounted for:

```php
if (count(array_filter([$abstractArg, $concreteArg, $sharedArg])) !== count($node->getArgs())) {
    return null;
}
```

Half the lines, and it also covers a named argument that isn't a parameter at all.

One knock-on: `singleton(Foo::class, $fn, true)` is now skipped rather than rewritten to `singleton($fn)`. That third argument is inert (PHP ignores extra positional args to a userland method), so the old rewrite was valid — bailing is just the conservative reading. Happy to special-case it back if you'd rather keep the coverage. Both shapes are in `skip_unmatched_arguments.php.inc`.

PR #568, opened 2026-09-14
Fix silent model corruption in TablePropertyToTableAttributeRector

edzus:
`TablePropertyToTableAttributeRector` rebuilds the `#[Table]` attribute from scratch whenever it
migrates a property. Two bugs in that rebuild change how the model behaves:

```php
// before
#[Table(name: 'users', timestamps: false)]
class User extends Model
{
    protected $primaryKey = null;
}

// after
#[Table(name: 'users', key: null)]
class User extends Model {}
```

1. **`timestamps:` and `dateFormat:` are dropped.** The rebuild reads back only `name`, `key`,
`keyType` and `incrementing`, so the other two arguments are lost. A model that loses
`timestamps: false` starts writing `created_at` / `updated_at` to a table that has no such columns.

2. **`$primaryKey = null` becomes `key: null`.** Laravel ignores that argument —
`initializeModelAttributes()` guards with `$table->key !== null` — so the model falls back to `'id'`
and `save()` / `delete()` match zero rows silently. `$primaryKey = null` is the only way to declare
a model with no single-column key, and `#[Table]` has no argument for it, so the property has to
stay put.

### Changes

- Read `timestamps` and `dateFormat` back when rebuilding the attribute.
- Treat a `null` `$primaryKey` as non-convertible: keep the property, emit no `key`.
- Three fixtures covering both.

The two fixes are independent, in separate commits.


calebdw:
> $primaryKey = null becomes key: null. Laravel ignores that argument —

can you please open a framework issue / PR for this? Seems like that should be fixed rather than leaving a property...might can pass `false` to `$key` instead when disabled

edzus:
> > $primaryKey = null becomes key: null. Laravel ignores that argument —
> 
> can you please open a framework issue / PR for this? Seems like that should be fixed rather than leaving a property...might can pass `false` to `$key` instead when disabled

@calebdw I agree the attribute should probably cover this at some point. But this rule is pinned to `laravel/framework >=13.0` so if `#[Table(key: false)` (or a new attribute which I would prefer instead) ships in `13.N` it's an error for anyone below that. So keeping the property seems like the right call either way.

About the failing PHPStan - it's in a file my changes does not touch, seems it's caused by php-parser dependency drift.

